Issue #30, opened 2021-09-02
Being able to search for text in names/descriptions and make it the default search type, ideally moving to one search type


robvanderveer:

robvanderveer:
Use case: As a user I would like to be able to be pointed to resources around a certain aspect by searching for that aspect using a text snippet, such as crypto, or session or json, or xxe, or injection. Preferably case-insensitive.
Instead of adding an extra search type to "standard" and "cre" we could easily remove the search types and just have a search button:
* If the text is of format 000-000 we search for cre.
* If the text starts with "standard:" we search for that standard (using partial search - because currently you need to get the exact name right and you want to be able to search for "standard:top" to get "Top10 2017")
* else we search in names and descriptions, and maybe in hyperlinks if that's possible.

robvanderveer:
This has been overtaken through other issues

PR #37, opened 2021-09-05
Full text search

### LINKED ISSUE #36: make all search case insensitive

northdpole:
closes  #36 and #30


northdpole:
This PR implements full text search. It also makes all searching in the database case insensitive which was a long-standing improvement

Formats supported:
        * 'CRE:<id>' will search for the <id> in cre ids
        * 'CRE:<name>' will search for the <name> in cre names
        * 'Standard:<name>[:<section>:subsection]' will search for
              all entries of <name> and optionally, section/subsection
        * '\d\d\d-\d\d\d' (two sets of 3 digits) will first try to match
                           CRE ids before it performs a free text search
        Anything else will be a case insensitive LIKE query in the database

Issue #14, opened 2021-08-30
Integrate frontend building to instructions and dockerfile


northdpole:

robvanderveer:
status?

northdpole:
closed by #64

northdpole:
Done

PR #64, opened 2021-09-11
add makefile, add contributing instructions and readme improvements

### LINKED ISSUE #48: Setup collaboration/feedback options for the community
Spyros I know that you have specific good ideas about this. This is my stab at what I think is required: At launch it's important for the community to share input and provide suggestions immediately, particularly for the data. It requires that the source data is provided in such a way that people can find the data that they have suggestions for and suggest changes. 
Processing changes directly to production is not high priority I guess, but registering input and comments is important.

northdpole:
closes #48 , #14

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,25 @@
+---
+name: New Issue Template
+about: 'Please use this template for all new issues. '
+title: ''
+labels: ''
+assignees: ''
+
+---
+## Issue
+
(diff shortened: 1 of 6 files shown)


robvanderveer (approved):
It looks good. Let's discuss how we can split the instructions in two parts: one for suggestions regarding the data, one for the app. Plus I believe we need a general high level design document that explains the setup and the whole idea. For a new person this is not apparent.

Meess (approved):

PR #118, opened 2021-10-08
uploaded cre explained pdf

robvanderveer:


northdpole:
updated the same pdf in the original place, no need for this anymore

PR #120, opened 2021-10-09
Visual fix for landing page not spanning full height on large views

### LINKED ISSUE #114: Still white block at bottom on large screens
### **What is the issue?**

The landing page has a white block at the bottom on large screens (chrome)

### Expected Behaviour

The bottom should be completely blue

### Actual Behaviour

See issue

### Steps to reproduce

Chrome on a 2K display full screen. Perhaps zoom out a bit. It should be easy to reproduce.

sandagolcea:
#### What does this PR do?

Landing page was not stretching the full width of the page on large view, causing a white box to appear after the page content. This is now fixed.
(fixes #114 )

### DIFF
--- a/application/frontend/src/app.scss
+++ b/application/frontend/src/app.scss
@@ -19,13 +19,11 @@ body {
   background-color: white;
 }
 
-.mount {
-  display: flex;
-  justify-content: flex-start;
-  align-items: stretch;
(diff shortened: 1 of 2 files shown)


northdpole:
Thank you for fixing this!

PR #124, opened 2021-11-04
Add definitions for Open Security Information Base (OSIB)

sslHello:
Hi @northdpole 
I've added a PR with definitions for import and export from OSIB YAML files. This includes a YAML file with some examples.

### DIFF
--- a/application/defs/include/osib_example.yml
+++ b/application/defs/include/osib_example.yml
@@ -1,0 +1,361 @@
+####################################################################################
+# osib_short.yaml: Open Security Information Base (OBIB), (c) OWASP Top 10 team    #
+# This work is licendes under a Creative Commons Attribution 3.0 Unported Licence  #
+# ---------------------------------------------------------------------------------#
+# Example File                                                                     #
+####################################################################################
+
+doctype:        OSIB
+schema:         90         # V 0.90
+date:           20211104
+children:
+  2:
+    alias:      "owasp"
+  "owasp":
+    attributes:
+      sources_i18n:
+        en:
+          name:             "Open Web Application Security Project&reg; (OWASP)"
+    children:
+      1:
+        alias:              "asvs"
+      "asvs":
+        attributes:
+          sources_i18n:
+            en:
+              name:         "OWASP Application Security Verification Standard (ASVS)"
+              source:       "https://owasp.org/www-project-application-security-verification-standard/"
+        children:
+          402:
+            attributes:
+              source_id: "V4.0.2"
+              sources_i18n:
+                en:
+                  name:     "OWASP Application Security Verification Standard (ASVS)-V.4.0.2"
+                  source:   "https://github.com/OWASP/ASVS/tree/v4.0.2/4.0"
+            children:
+              1: # 1
+                attributes:
+                  source_id: "V1"
+                  sources_i18n:
+                    en:
+                      name:     "V1-Architecture, Design and Threat Modeling Requirements"
+                      source:   "https://github.com/OWASP/ASVS/blob/v4.0.2/4.0/en/0x10-V1-Architecture.md"
+                  links:
+                  - link: "osib.owasp.asvs.402"
+                    type: "parent"
+                  - link: "osib.owasp.asvs.402.1.1"
+                    type: "child"
+                  - link: "osib.owasp.asvs.402.1.2"
+                    type: "child"
+                  - link: "osib.owasp.asvs.402.1.4"
+                    type: "child"
+                  - link: "osib.owasp.asvs.402.1.5"
+                    type: "child"
+                  - link: "osib.owasp.asvs.402.1.6"
+                    type: "child"
+                  - link: "osib.owasp.asvs.402.1.7"
+                    type: "child"
+                  - link: "osib.owasp.asvs.402.1.8"
+                    type: "child"
+                  - link: "osib.owasp.asvs.402.1.9"
+                    type: "child"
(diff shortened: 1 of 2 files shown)


northdpole:
hi, commenting to say, i'm still working on this, it's significantly sized and requires lots of thought

application/defs/include/osib_example.yml
```diff
@@ -0,0 +1,361 @@
...
+# This work is licendes under a Creative Commons Attribution 3.0 Unported Licence  #
+# ---------------------------------------------------------------------------------#
+# Example File                                                                     #
+####################################################################################
+
+doctype:        OSIB
+schema:         90         # V 0.90
+date:           20211104
+children:
+  2:
```
northdpole, line 12: hey @sslHello if we have multiple children this needs to be a list, right?
so the line should be the following

```suggestion
-  2:
```

application/defs/include/osib_example.yml
```diff
@@ -0,0 +1,361 @@
...
+    alias:      "owasp"
+  "owasp":
```
northdpole, line 14: does "owasp" need to be it's own child element?
can it be part of element "2"?
if yes, then do we need this line?
``` "owasp":  ```` 
we already know the element's alias.

File: application/defs/include/osib_example.yml [hunk 3]
```diff
@@ -0,0 +1,361 @@
...
+    attributes:
+      sources_i18n:
+        en:
+          name:             "Open Web Application Security Project&reg; (OWASP)"
+    children:
+      1:
```
northdpole, line 20: can you remind me why we need numbered children? could we instead force the aliases to be unique ids or names?
```suggestion
      - id: 1
 ```
or
```suggestion
      - name: 1
```

application/defs/include/osib_example.yml
sslHello: As for now a children are nested dictionaries right now, same for 'sources_i18n'. 
In my opinion, it is easier and more efficient to adress items directly using a directory than a list. Lists are a kind of step in the dark, as you don't know the items and you have to lookup them item by item to find the right element...
How about using dictionaries for nodes and elements that can ge precisely addess by a path e.g. osib.owasp.top10.2021-1-0.1 = Top10, 2021, V1.0, A01:2021?

application/defs/include/osib_example.yml
sslHello: Yes, I've histirically started with a numbering path for all nodes, as YAML supports names, I'd prefer names as they are more human readable ;-). The alias use case makes more sense for 'aliasing' versions that do not differ in their osib representation. But i's only an extension (a gimmick), no core element of the schema, if you define identical (non breaking) (major) versions and ignore subversions, e.g. asvs V4.0.x -> osib.owasp.asvs.4-0
By the way we sould define a convention how to name versions. For Performance reasons dots ('.') should be avoided, so I use dashes '-' right now.Please finf enclosed the latest OSIB yaml file I use for testing the Python scripts for the two MkDocs macros: https://github.com/sslHello/OSIB-Test/blob/main/osib_short.yml

application/defs/include/osib_example.yml
sslHello: Yes, we don't need numbers for the projects, standards or tools. They were there for historical reasons or if we like to define a stable 'Id' for a context root. Names can change...
Other possible solution: use the 'alias' token to accept changing names for projects, standards or tools and to provide backward compatibility...

sslHello: We will need numbers inside the project's, standard's or tool's subtrees as theiry table of contents use then usually to identify chapters, sections and subsections.

application/defs/include/osib_example.yml
northdpole: > In my opinion, it is easier and more efficient to address items directly using a directory than a list. 
Fair, will keep it a Dict[str, Osib_node] then

application/defs/include/osib_example.yml
northdpole: > For Performance reasons dots ('.') should be avoided, 
Can you please elaborate on this? Semantic versioning uses dots

sslHello: > >  Yes, I've historically started with a numbering path for all nodes, as YAML supports names, I'd prefer names as they are more human readable ;-).
> understood. I'll make the "2" be a string then
> **But** I will get rid of the extra "owasp" on line 14 as it duplicates info we already know about
I'd suggest to do just the other way round: use 'owasp' as key for rge ditionary and get rid of the '2', OK?

sslHello: >>   For Performance reasons dots ('.') should be avoided,

> Can you please elaborate on this? Semantic versioning uses dots#

We do need one subtree by version to manage versions. To do this, we add ".<version>" to the path of any context root of any project, standard or tool. This adds "children: { <version>: }" to the tree => adds 1 level of tree depth, assuming that 'children' has not been the sole attribute of the project, standard or tool (e.g. name, source_id etc).
If we use doths inside the version, this works but generates additional layers of depth without any further information:
* Version ".1.0" -> adds "children: {1: { children: { 0: } } }" => 3 layers (= 2 empty layers more than above)
* Version ".1.0.0" -> adds "children: {1: { children: { 0: { children: {0: } } } } }" => 5 layers (= 4 empty layers more than above)
The yaml tree gets blown up and any search/lookup needs to search through these additional layers...
The schema is able to handle both, but a convention *not* to use dots ('.') as seperator in versions produces a taller YAML tree.

PR #125, opened 2021-11-13
Filtering experiment

northdpole:
testing instructions: 
terminal: cp cres/db.sqlite standards_cache.sqlite;  yarn start; make dev-run

in a browser: 
localhost:9001/standard/ASVS or localhost:9001/search/asvs
expand dropdowns and click on the "Filter" button
then click apply filter on the header bar
you should be able to see the links to the CRE-ids you have selected

the clear bytton should clear the url and re-show everything


application/frontend/src/scaffolding/Header/Header.tsx
```diff
@@ -15,7 +15,24 @@ const getLinks = (): { to: string; name: string }[] => [
 ];
 
 export const Header = () => {
-  const { params, url, showHeader } = useLocationFromOutsideRoute();
+const HandleDoFilter = () => {
+    currentUrlParams.set("applyFilters", "true");
+    history.push(window.location.pathname + "?" + currentUrlParams.toString());
+  }
+
+  const ClearFilter = () => {
```
sandagolcea, line 23: Normally functions are lowerCamelCase in js.
When you have `export const SomeComponent`, you use PascalCase to denote that is a React component.

TL;DR: naming convention is lowerCamelCase for functions , PascalCase for React components

application/frontend/src/components/DocumentNode/DocumentNode.tsx
```diff
@@ -117,9 +134,19 @@ export const DocumentNode: FunctionComponent<DocumentNode> = ({ node, linkType,
...
-                        <DocumentNode node={link.document} linkType={type} hasLinktypeRelatedParent={isNestedInRelated()} key={i} />
-                      )
+                    {links.map((link, i) =>
+                      (!(currentUrlParams.has("filters")) ||
+                        !currentUrlParams.has("applyFilters") ||
+                        currentUrlParams.get("applyFilters") == "false") ||
+                        (currentUrlParams.get("applyFilters") == "true" &&
+                          currentUrlParams.has("filters") &&
+                          link.document.id &&
+                          currentUrlParams.getAll("filters").includes(link.document.id)) ?
```
sandagolcea, line 144: You can use a shorthand instead of the ternary here: 
 i.e.replace the ternary: `condition ? <div/> : ""` 
with a shorthand i.e. `condition && <div/>`

application/frontend/src/components/DocumentNode/DocumentNode.tsx
```diff
@@ -61,54 +78,54 @@ export const DocumentNode: FunctionComponent<DocumentNode> = ({ node, linkType,
...
-      </>
-    
+      <span>
+        Reference:
+      </span>
+      <a href={hyperlink.hyperlink} target="_blank"> {hyperlink.hyperlink}</a>
+    </>
   }
-  
   const SimpleView = () => {
```
sandagolcea, line 102: if you have a component that has just a return, you can use some syntactic goodness, and make it shorter:

```
const SimpleView = () => {
  // no extra logic here
  return (<div></div>);
}

```
bit cleaner:


```
const SimpleView = () => (<div></div>);

```


This is common practice, usually when you can do this, you do.

PR #147, opened 2022-01-28
zap alerts to cre, init

northdpole:
depends on #129 

adds zap alerts (automatically) to cre
closes https://github.com/OWASP/common-requirement-enumeration/pull/147
todo:

- [ ] create github action that runs `python cre.py --zap_in  ` periodically and updates both cres/db.sqlite and heroku database


northdpole:
closed in #163  instead

PR #156, opened 2022-02-18
Cs parse

northdpole:
parse CRE links in the owasp cheatsheeets that exist either in comments or normally and add them to the graph


northdpole:
closed in #164

PR #274, opened 2023-04-25
Changes to support he new higher level structure

northdpole:


application/utils/external_project_parsers/juiceshop.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        return
+    challenges = yaml.safe_load(resp.text)
+    for challenge in challenges:
+        chal = defs.Tool(
+            description=challenge["description"],
+            name="OWASP Juiceshop",
+            section=challenge["name"],
+            sectionID=challenge["key"],
+            hyperlink="https://preview.owasp-juice.shop/#/score-board?challenge="
```
bkimminich, line 39: You are pulling the challenges from the `master` branch but then link to the environment built from `develop`. It would recommend to link to https://demo.owasp-juice.shop instead as that is consistent with `master`.

application/utils/external_project_parsers/juiceshop.py
```diff
@@ -0,0 +1,80 @@
...
+    if resp.status_code != 200:
+        logger.fatal(
+            f"could not retrieve challenges yaml, status code {resp.status_code}"
+        )
+        return
+    challenges = yaml.safe_load(resp.text)
+    for challenge in challenges:
+        chal = defs.Tool(
+            description=challenge["description"],
+            name="OWASP Juiceshop",
```
bkimminich, line 36: Project name is spelled "Juice Shop".

PR #291, opened 2023-05-29
Renamed 'is same as' to 'is linked using AI to'

robvanderveer:


northdpole:
i'm not sure I understand what's the use case for this?
is the technical detail that we use LLMs for some of our mappings important enough to warrant us repeating it site-wide?

robvanderveer:
The problem this change addresses is that the current 'is the same as' is a wrong description. For the normal links we have the description of 'is linked to'. The 'is the same as' is a much stronger claim, but the irony is that the link is much weaker because it's either through transitive linking or a best guess using LLM. We need a way to name this group of links aptly and also make clear that it is typically of lower quality. Hence the 'linked using AI' idea. I'm open to alternatives, as long as we solve the problem.

northdpole:
Thanks @robvanderveer for the contribution and for putting time into OpenCRE — please keep contributing when you can.

**Closing:** Superseded on main: link type is already `Automatically linked to` / `has been automatically mapped to` (evolved wording of this rename).

PR #318, opened 2023-07-04
Standardize document type text selection

john681611:
This solution uses the first doctype to identify if its a CRE or not to do the switchover but of course, there is a risk of mixed doc types.

We can add that to the grouping logic if we want.


northdpole:
is this ready to be merged?

PR #377, opened 2023-09-14
Neo4j standard object

john681611:


application/database/db.py
```diff
@@ -224,32 +224,73 @@ def add_cre(self, dbcre: CRE):
...
+            nid=dbcre.id,
             description=dbcre.description,
-            external_id=dbcre.external_id,
+            links=[],  # dbcre.links,
+            tags=dbcre.tags,
+            metadata="{}",  # dbcre.metadata,
             database_="neo4j",
         )
 
     @classmethod
```
northdpole, line 238: is this necessary to be a classmethod?

application/database/db.py
```diff
@@ -224,32 +224,73 @@ def add_cre(self, dbcre: CRE):
...
-            subsection=dbnode.subsection or "",
-            tags=dbnode.tags,
-            version=dbnode.version or "",
-            description=dbnode.description,
-            ntype=dbnode.ntype,
-            database_="neo4j",
-        )
+        if dbnode.ntype == "Standard":
+            self.driver.execute_query(
+                "MERGE (n:Standard {id: $nid, name: $name, section: $section, sectionID: $sectionID, subsection: $subsection, tags: $tags, version: $version, description: $description, doctype: $doctype, links: $links, metadata: $metadata, hyperlink: $hyperlink})",
```
northdpole, line 244: can we use sqlalchemy here? i'd prefer it if we used an ORM, if it's impossible its ok

application/database/db.py
```diff
@@ -289,7 +330,7 @@ def gap_analysis(self, name_1, name_2):
             return None, None
         base_standard, _, _ = self.driver.execute_query(
             """
-            MATCH (BaseStandard:Node {name: $name1})
+            MATCH (BaseStandard:Standard|Tool {name: $name1})
```
northdpole, line 333: what is a base standard?

application/database/db.py
```diff
@@ -325,70 +366,94 @@ def gap_analysis(self, name_1, name_2):
...
 
+    @classmethod
+    def parse_node(self, node: neo4j.graph.Node) -> cre_defs.Document:
+        name = node["name"]
+        id = node["id"] if "id" in node else None
+        description = node["description"] if "description" in node else None
+        # links = [self.parse_link(link) for link in node["links"]]
+        tags = node["tags"]
+        # metadata = node["metadata"]
+        if "Code" in node.labels:
```
northdpole, line 403: please do not hardcode the node types here, instead it should be `if cre_defs.doctypes.Code.value in node.labels:` this way we can change the doctypes easily without having to match and replace

application/database/db.py
john681611: Copying this in terms of naming one is a base to compare against (if we have more compares)
Screenshot 2023-09-15 at 15 39 54

application/database/db.py
john681611: Can we make this a separate issue that must be handled before we add more queries?  It seems more suitable as a separate PR.

application/database/db.py
john681611: Made static

Issue #374, opened 2023-09-13
Chatbot answers miss formatting


robvanderveer:
Chatbot answers seem to have some layout with bullets etc but we don't render it as bullets. Could we?

northdpole:
it would require us to trust whatever comes from google, given that xss via markdown is a valid attack pattern i'm not willing to do that  and I haven't managed to find a better way to do this yet

PR #384, opened 2023-09-17
make returned markdown both pretty and safe

northdpole:
while also providing decent syntax highlight

current bug: dompurify removes the code to be highlighted


northdpole:
closes #374 , thanks to Sam Stepanyan for the technical pointers

northdpole:
image

PR #447, opened 2023-11-13
Added SDK

stateandprove:
I've created an SDK for Open CRE API.


northdpole:
Thank you for the SDK contribution! We're going in a different direction with the MCP server (#1013) and the existing REST API. Closing with appreciation.

PR #467, opened 2024-01-07
Switch to new challenge search syntax

bkimminich:
Since v16.x the new Score Board with its generic `searchQuery=` parameter is available and the default in Juice Shop. While `challenge=` is still working as an alias for now, it will likely be removed in release v17.x. See here for new syntax: https://pwning.owasp-juice.shop/companion-guide/latest/part4/integration.html#_generating_links_to_juice_shop


northdpole (approved):
thanks @bkimminich !

Issue #507, opened 2024-05-16
Make standard tags in explorer link directly to the right entry/entries for the CRE


robvanderveer:
The Explorer now shows buttons for the standards in rows. Currently that button takes you to the entire standard. Instead, please let it link to the deeplink (if available) to the content of that standard, for the CRE in that row. 
Sometimes these are multiple standards (like often with Capec): that could take the user to an OpenCRE page that shows the CRE and just filters on CAPEC. So then the user can choose which capec link to go to.

Am I correct in assigning this to you Diana? :)

dlicheva:
Thanks @robvanderveer , this should be done here: https://github.com/OWASP/OpenCRE/pull/513
Could you have a look to see if that's the right idea please?

PR #513, opened 2024-05-26
[507] UI-only: Link standard tags in /explorer directly to the right entries for the CRE

dlicheva:
[Ticket](https://github.com/OWASP/OpenCRE/issues/507)
This PR makes it so that the standards on the /explorer page now link to the filtered CRE page filtered by them.
For example:
1. A CRE has links to all of these standards: 
Screenshot from 2024-06-08 18-19-23
2. You click on NIST 
Screenshot from 2024-06-08 18-20-41
3. You get the CRE page showing the NIST sections that are related to that CRE: 
Screenshot from 2024-05-26 21-57-47

4. If the link is to an ungrouped standard and has an external link, you would be linked directly to that page:
Screenshot from 2024-06-08 18-20-00
Screenshot from 2024-06-08 18-20-22

### DIFF
--- a/application/frontend/src/components/FilterButton/FilterButton.tsx
+++ b/application/frontend/src/components/FilterButton/FilterButton.tsx
@@ -1,13 +1,8 @@
-import axios from 'axios';
-import React, { FunctionComponent, useEffect, useMemo, useState } from 'react';
-import { Link, useHistory } from 'react-router-dom';
+import React, { FunctionComponent, useState } from 'react';
+import { useHistory } from 'react-router-dom';
 import { Button } from 'semantic-ui-react';
 
-import { useEnvironment } from '../../hooks';
-import { Document, LinkedDocument } from '../../types';
-import { getDocumentDisplayName, groupLinksByType } from '../../utils';
-import { getApiEndpoint, getInternalUrl } from '../../utils/document';
-import { LoadingAndErrorIndicator } from '../LoadingAndErrorIndicator';
+import { Document } from '../../types';
 
 export interface FilterButton {
   document: Document;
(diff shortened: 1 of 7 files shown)


robvanderveer:
This is wonderful!
The idea was that if there is only one resource to jump straight to the hyperlink (if any) of that resource. So if there's only one ASVS entry then clicking the 'tag' will take the user to the ASVS page outside of OpenCRE. If that's not in place yet, no worry; it can always be a second step.

dlicheva:
Hi @robvanderveer, I can do that, but wouldn't it feel inconsistent to the user? e.g. some tags would lead to a CRE page and others would lead to an external page. Once users are out of the website it can be confusing to navigate back. 

Alternatively, if there is only one link, the tag can link to the content on OpenCRE.org? e.g. linking directly to https://opencre.org/node/standard/NIST%20800-53%20v5/section/CM-2%20BASELINE%20CONFIGURATION ? Then the user can navigate to the external resource, through a more obvious link.

What do you think?

dlicheva:
Thanks @robvanderveer, I see ! I think opening the deeper linked standard in a new tab as you suggest, would be the best approach :+1: 
This should be less confusing because openCRE.org/explorer is there for the user to return to, and they get a new tab with the content of the standard.

dlicheva:
Hi @robvanderveer , this is updated now, so if there is a link with a hyperlink and no siblings, you would get a new tab on click. Is this ok?:
[Screencast from 08-06-24 18:22:50.webm](https://github.com/OWASP/OpenCRE/assets/8710269/b7bf397b-51d7-4f55-b657-7d15b61a61d9)

robvanderveer:
Yup that looks amazing.

northdpole (approved):

Issue #514, opened 2024-06-11
Explorer is incomplete


robvanderveer:
opencre.com/explorer is missing a lot of branches, for example everything under input and output protection.
If adding those branches makes building the tree up slower, the build up probably will need to be optimized.

northdpole:
screenshot of the problem, data is there, but the presentation layer doesn't loop enough (should be recursion maybe?)
![image](https://github.com/OWASP/OpenCRE/assets/1583262/9f7bf5f4-2f31-4c52-ad0d-c960449387b3)

dlicheva:
Hi @northdpole ,
The data is in the dataStore (for the force graph), but is not returned by the api when calling the /root_cre endpoint. The explorer uses that endpoint and not the dataStore. I'll keep looking, but maybe you understand the back-end code better, so it's worth a look.
![Screenshot from 2024-06-22 18-13-56](https://github.com/OWASP/OpenCRE/assets/8710269/a0965c4c-ed61-4e56-a759-b2f62fe6a011)

Wassaf001:
Hey, what is the issue?

Can anyone explain, would love to work on it.

PR #519, opened 2024-06-27
Bugfix/514 🐛 Fix explorer incomplete

dlicheva:
This PR is a step further on the issue [here](https://github.com/OWASP/OpenCRE/issues/514), trying to fix the issue with the incomplete explorer.
My previous PR was only including one level of links once it reached the end of the root_cres response: https://github.com/OWASP/OpenCRE/pull/518

This PR includes the full tree.
However, the full tree doesn't fit into local storage, so the page loads really slowly.
It might be ok for a proof of concept, but requires further discussion.
Screenshot from 2024-06-27 15-40-11

If this is merged, you'll need to clear your application cache and refresh the page. Local storage lasts 2 days, so users who don't know this can see the new content in two days.


dlicheva:
I'll use the change here to get the CREs all at once from the API: https://github.com/OWASP/OpenCRE/pull/541

PR #521, opened 2024-07-06
design for my opencre

northdpole:


robvanderveer (approved):
The story is correct but sends an unnecessary negative image of OpenCRE as if it was badly designed. Every platform stage has its own reasons and history. It is phrased as something that users cannot. Instead, we can phrase it as something that users want and for which we will build support.

It says data and policies. That should be data.

Export cre structure as CSV -> Export cre structure as mapping template to contribute mappings between the cre catalog and a standard in CSV format

Given a CSV describing a standard -> Given a CSV containing a standard with a requirement in each row

to suggest the nearest cre -> to match the nearest cre for every requirement in the standard and populate the mapping template

Given a CSV mapping cre to standard clauses, import the standard and calculate gap analysis.
->
Given a populated mapping template and a CSV containing a standard, import the standard and calculate gap analysis

Add:
* Document that the mapping template may be used to suggest changes in the OpenCRE catalog and how (name changes, moving items in the catalog, adding requirements, adding requirement categories. For now, these suggestions will be handled manually.

Please update the text in the further paragraphs in line with the above changes (I'm happy to do that myself)

By the way, it would be good to just commit this design so we can edit it. This review workflow is not ideal, right? If it is commited, I'll have a further look at details.

Issue #531, opened 2024-07-21
Allow users to download CSV with all CREs in offset format


northdpole:
### **What is the issue?**

The first step of [MyOpenCRE](https://github.com/OWASP/OpenCRE/blob/main/docs/designs/my-opencre.md) is an endpoint that allows users to download an exported csv of all CREs in the current instance
This ticket tracks the creation of such an endpoint

PR #532, opened 2024-07-21
fix the api part of #531 by providing an api call to export all CREs …

northdpole:
…in an instance in a visually pleasing format

This produces spreadsheets like this one https://docs.google.com/spreadsheets/d/10jrVWRRw-zCVFVgzumnO7xgNTjZYU6M6B9EDqss9zxY/edit?usp=sharing

Todo:
- [x] adjust headers
- [x] ensure data consistency

### DIFF
--- a/application/tests/spreadsheet_test.py
+++ b/application/tests/spreadsheet_test.py
@@ -4,7 +4,10 @@ import unittest
 from application import create_app, sqla  # type: ignore
 from application.database import db
 from application.defs import cre_defs as defs
-from application.utils.spreadsheet import prepare_spreadsheet
+from application.utils.spreadsheet import (
+    prepare_spreadsheet,
+    generate_mapping_template_file,
+)
 
 
 class TestDB(unittest.TestCase):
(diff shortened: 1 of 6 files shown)


robvanderveer:
The example sheet: https://docs.google.com/spreadsheets/d/1btYiu95x7n3BbSy47xx-YMf2IBMaMX5KdrEBFsNmQzw/edit?gid=0#gid=0

application/tests/spreadsheet_test.py
```diff
@@ -4,7 +4,7 @@
 from application import create_app, sqla  # type: ignore
 from application.database import db
 from application.defs import cre_defs as defs
-from application.utils.spreadsheet import prepare_spreadsheet
+from application.utils.spreadsheet import prepare_spreadsheet, generate_staggered_csv
```
robvanderveer, line 7: shouldn't we use more functional naming? so instead of staggered_csv say mappingtemplatefile?

application/utils/spreadsheet.py
```diff
@@ -237,3 +237,59 @@ def write_spreadsheet(title: str, docs: List[Dict[str, Any]], emails: List[str])
...
+                    rows.extend(
+                        add_offset_cre(
+                            cre=link.document,
+                            database=database,
+                            offset=offset + 1,
+                            visited_cres=visited_cres,
+                        )
+                    )
+                elif link.ltype == defs.LinkTypes.Related:
+                    related.add(link.document.id)
```
robvanderveer, line 274: Currently this code is doing nothing with related CRES, but we need to have them in the template at some point, for the template to be usable for changing the catalog.  We have two options: add them as id|name list in one extra column, or add them as rows, just like we're adding the children (contains) as rows.  I think the latter is easier to read but more confusing. So I vote for one column for the related CRES. This would require building up a dictionary of the related id|name pairs, and then adding them to rows, together with the CRE in a dictionary of two (or one if there is no related CRES.

application/utils/spreadsheet.py
```diff
@@ -237,3 +237,59 @@ def write_spreadsheet(title: str, docs: List[Dict[str, Any]], emails: List[str])
...
+    related = set()
+
+    def add_offset_cre(
+        cre: defs.CRE, database: db.Node_collection, offset: int, visited_cres: Set
+    ) -> List[Dict[str, str]]:
+        nonlocal maxOffset, related
+        maxOffset = max(maxOffset, offset)
+        rows = []
+
+        rows.append({f"CRE{offset}-ID": cre.id})
```
robvanderveer, line 253: We want to get rid of the -ID and do f"{id | name}|:cre.id, cre.name} instead of cre.id. Right?

robvanderveer (commented):
I leave it to you Spyros what to do with the comments. This code looks like it will work, but it needs some changing to have the functionality we want.

application/utils/spreadsheet.py
```diff
@@ -237,3 +237,59 @@ def write_spreadsheet(title: str, docs: List[Dict[str, Any]], emails: List[str])
...
+        visited_cres.add(cre.id)
+        dbcre = database.get_CREs(external_id=cre.id)
+        if not dbcre:
```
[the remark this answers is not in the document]
northdpole: dbcre should be None in case of error

northdpole:
https://docs.google.com/spreadsheets/d/1btYiu95x7n3BbSy47xx-YMf2IBMaMX5KdrEBFsNmQzw/edit?gid=1795535977#gid=1795535977   <-- latest import

application/utils/spreadsheet.py
northdpole: > Currently this code is doing nothing with related CRES, but we need to have them in the template at some point, for the template to be usable for changing the catalog. 

I agree

>  We have two options: add them as id|name list in one extra column, or add them as rows, just like we're adding the children (contains) as rows

We could have a column named "Related CRE" sure, 
then we can add the CRE listed in this column with a related type link,
During importing tehn we have the hierarchy on the left, then one column with related CREs, then an arbitrary number of standard entries. Do we link each standard to both the normal cre and the related one? This seems  a bit confusing for people. 

Since we want to be capturing more complex relationships in a 2 dimensional matrix I think we should consider a more verbose and less context-based format.

e.g.
* perhaps there should be 2 formats, 1 for mapping standards and one for links between cres
* alternatively perhaps everything should be considered a standard, CREs-included, we forget the hierarchy and just support a link between only 2 standards for each CSV, these 2 standards are CRE and any standard, for this we need to explicitly include the relationship type in a column.

In any case, this is beyond the scope of this particular pull request, let's think about this a bit more, let me know when you are free for a brainstorming session

robvanderveer: The mapping template specifies with which CRE to link standard entries. The related CREs will not get the standard entries aside. I see your point. I guess it can be an option that is off by default to include the related CREs column in the template. So yes: two formats. and indeed not for this PR.

application/utils/spreadsheet.py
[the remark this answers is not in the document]
robvanderveer: Does not None resolve to true?

PR #539, opened 2024-07-29
Removed: opentelemetry

john681611:
Request from @northdpole


northdpole (approved):
thank you, great work!

PR #547, opened 2024-08-17
Suggest mappings

northdpole:
This pr introduces a new endpoint which accepts a csv in the import format with lines that do not have a CRE and uses embeddings to fill the missing cres with a suggestion.

- [ ] add ability to get  matches by defining a prompt (hopefully better accuracy)
- [ ] add ability to define openai endpoint so that it can interface with other ai engines that are openai compatible
- [ ] accurately populate cre cardinality on the spreadsheet

You can test this pr as follows:

```
# sync from upstream
python cre.py --upstream_sync  

# generate embeddings, WARNING if you want any sort of accuracy you should use an LLM instead of spacy which is for local unit tests only
python cre.py --generate_embeddings 

# run the web server
make dev-flask

# on another terminal
curl -X POST http://localhost:5000/rest/v1/cre_csv/suggest -F "cre_csv=@< csv file that follows the export format but does not necessarily have cres>" 
```

### DIFF
--- a/application/prompt_client/prompt_client.py
+++ b/application/prompt_client/prompt_client.py
@@ -1,6 +1,10 @@
 from application.database import db
-from application.defs import cre_defs
-from application.prompt_client import openai_prompt_client, vertex_prompt_client
+from application.defs import cre_defs as defs
+from application.prompt_client import (
+    openai_prompt_client,
+    vertex_prompt_client,
+    spacy_prompt_client,
+)
 from datetime import datetime
 from multiprocessing import Pool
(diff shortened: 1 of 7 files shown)


northdpole:
Thanks @northdpole for the contribution and for putting time into OpenCRE — please keep contributing when you can.

**Closing:** Closing unfinished embeddings CSV 'suggest mappings' experiment; direction moved on (MyOpenCRE / Librarian). Please open a fresh PR if you want to revive a specific piece.

PR #590, opened 2025-03-06
Change in requirement.txt and Makefile for Linux installation

Pa04rth:
During the OpenCRE installation over Linux , I encountered some errors which after discussing with the maintainer @northdpole  were solved ,below is the detailed explanation about each error and how I solved it :

Documentation Used : [developmentSetup](https://github.com/OWASP/OpenCRE/blob/37d5db34ee1434421b968982dce8474c1cf047c0/docs/developmentSetup.md)

1. After cloning the forked repo , when I entered the `make install` command ,it showed me the error - Building wheel for typed_ast failed.
image
Then to solved this I removed the typed_ast from requirement.txt file {As this file is of no importance in current versions}
2. Now , after solving the above error a new error came related to system-site-packages -
image
The above error was solved by removing the system-site-packages line from "install python" function in Make file -
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ install-deps-typescript:
 install-deps: install-deps-python install-deps-typescript
 
 install-python:
-	virtualenv -p python3 --system-site-packages venv
+	virtualenv -p python3  venv
 	. ./venv/bin/activate &&\
 	make install-deps-python &&\
 	playwright install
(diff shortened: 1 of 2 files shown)


northdpole (approved):

Pa04rth:
Sorry for the extra commits ,actually I want to add the DevelopSetup doc commit to a newer PR ,but by mistake I did on same ,So I need to revert it .

northdpole (approved):

northdpole (approved):

Issue #486, opened 2024-02-08
Make smartlink go to CRE directly


robvanderveer:
If you smartlink for example with CWE 611 and there's only one CRE connected, it should jump straight to that CRE and show the wealth of information. That CRE page has all the information that the current smartlink landing page has, so.
https://www.opencre.org/smartlink/standard/CWE/611](https://www.opencre.org/smartlink/standard/CWE/611

shreeshtripurwarcomp23-coder:
Hi, @northdpole, @robvanderveer, I’d like to work on this issue if it’s still available.

From my understanding, the goal is to redirect directly to the CRE page when the smartlink resolves to a single CRE result (for example CWE-611), instead of showing the intermediate landing page.

I’m currently exploring the codebase and would be happy to work on implementing this.

robvanderveer:
Go ahead.

shreeshtripurwarcomp23-coder:
I will ! Thank you for the opportunity

shreeshtripurwarcomp23-coder:
@robvanderveer @Pa04rth @manshusainishab @northdpole @Bornunique911 
Hi! I've submitted a PR for this issue:
https://github.com/OWASP/OpenCRE/pull/908

PR #597, opened 2025-04-02
Changes in web_main to redirect smartlinks directly to CRE (Issue #486)

Pa04rth:
Problem - The specific issue at hand is to modify the behavior of the `/smartlink` route so that if a smartlink is connected to only one CRE, it directly redirects to that CRE's page instead of landing on the generic smartlink page.

Solution - In my following commit , I have first check the number of nodes connected to that particular CRE and if it is one than I have changed the redirect url .

### DIFF
--- a/application/cmd/cre_main.py
+++ b/application/cmd/cre_main.py
@@ -335,12 +335,20 @@ def parse_standards_from_spreadsheeet(
                 register_cre(cre, collection)
                 bar()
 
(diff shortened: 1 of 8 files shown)


northdpole:
Thanks @Pa04rth for the contribution and for putting time into OpenCRE — please keep contributing when you can.

**Closing:** Superseded: smartlink already redirects to the single linked CRE on main; issue #486 is closed.

Issue #619, opened 2025-07-24
Remaining zoomable circles issues


robvanderveer:
The zoomable circles does not resize when the browser resizes. The green background does, but the circles don't. On Chrome on MacOS.
Rename ‘Cluster’ to ‘OpenCRE’
Zoom in out is in too small steps I think. Make it three times as big steps.

While playing around I learned we definitely need the labels.
Because they can get in the way we need to make them switchable and have them on by default.
How about we try to display them at the top of the circle? That will be more clear I think because then it’s clear it belongs to the circle. When its printed in the middle, that’s where other circles are.
Can you give this a try?

PR #620, opened 2025-07-24
Fix: Zoomable circles responsive resize, label toggle, and zoom steps (#619)

nashutosh:
🔧 Summary of Fixes for Issue #619
This PR addresses the zoomable circle visualization issues and improves overall responsiveness and usability.

✅ Fixes Implemented:
Responsive Resize: Circles now correctly resize with browser window changes by adding width and height to the dependency array in the useEffect.

Label Toggle: Added a checkbox to toggle labels (enabled by default) for better visibility and cleaner UI.

Label Positioning: Moved labels to the top of each circle using the dy attribute for better clarity (instead of center).

Zoom Improvement: Made zoom steps 3x larger for a more noticeable zoom effect by adjusting the margin calculation logic.

Renaming: Renamed the root node label from 'Cluster' to 'OpenCRE' for consistency with project terminology.

### DIFF
--- a/application/frontend/src/pages/Explorer/visuals/circles/circles.scss
+++ b/application/frontend/src/pages/Explorer/visuals/circles/circles.scss
@@ -16,9 +16,9 @@
     font: 11px "Helvetica Neue", Helvetica, Arial, sans-serif;
     text-anchor: middle;
     text-shadow: 0 1px 0 #fff, 1px 0 0 #fff, -1px 0 0 #fff, 0 -1px 0 #fff;
+    pointer-events: none;
 }
 
-.label,
 .node--root,
 .node--leaf {
     pointer-events: none;
@@ -29,4 +29,8 @@
     right: 0;
     margin: 0;
     background-color: transparent;
+}
+
+.ui.toggle.checkbox {
+    margin-right: 10px;
 }
--- a/application/frontend/src/pages/Explorer/visuals/circles/circles.tsx
+++ b/application/frontend/src/pages/Explorer/visuals/circles/circles.tsx
@@ -4,13 +4,15 @@ import { LoadingAndErrorIndicator } from 'application/frontend/src/components/Lo
 import useWindowDimensions from 'application/frontend/src/hooks/useWindowDimensions';
 import { useDataStore } from 'application/frontend/src/providers/DataProvider';
 import * as d3 from 'd3';
-import React, { useEffect, useState } from 'react';
-import { Button, Icon } from 'semantic-ui-react';
+import React, { useEffect, useState, useRef } from 'react';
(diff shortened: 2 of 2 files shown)


robvanderveer:
thanks! how do you think the circle labels look now? Any thoughts on how to improve labeling overall?

nashutosh:
Thanks! How do you think the circle labels look now? Any thoughts on how to improve labeling overall?
@robvanderveer

Hi Sir! Thanks for taking the time to review and raise the question — I think the circle labels are looking much better now. They’re clearly centered within each group, responsive across zoom levels, and readable without overlapping. The earlier issue with transform calculations and misaligned text was resolved by refining the coordinate logic.

That said, I’d love to hear your thoughts on how we can take labeling further:

We could explore [D3's textPath](https://github.com/d3/d3-shape#arcs) or SVG path-following for more dynamic placement.

Another idea is adding tooltips or hover cards to handle long or hidden labels at lower zoom levels.

If helpful, we can also fine-tune font scaling, add icons, or semantic color cues for better visual grouping.

Also — one small UI quirk I noticed: the footer currently blocks scrolling, which causes elements to get hidden behind it. I’ll take a look at that separately unless someone’s already working on it.

ℹ️ Just for context:
The data shown here is based on mock data for now, as I wasn’t able to run the backend locally due to a connection error. But the structural rendering and label logic should remain the same.

robvanderveer:
Thanks, but I don't understand the images. Aren't the labels supposed to be printed at the top of the circles?

nashutosh:
> Thanks, but I don't understand the images. Aren't the labels supposed to be printed at the top of the circles?

Hi Sir,

Thanks again for pointing that out!

Currently, I’m applying the label transform relative to the center of each circle using a basic translate(x, y - r - offset) logic — so the labels should ideally appear at the top of the circles.

Since I was testing with mock data (as the backend wasn't active), it's possible the radius or coordinate values weren’t accurate, which caused the label to misalign in the screenshot.

If you have a preferred approach for positioning — say, anchoring exactly at the top center using a consistent margin or switching to a textPath — I’d be happy to implement that. Just let me know what style or direction you’d like me to follow.

Thanks once again! Looking forward to refining this based on your feedback 🙏

nashutosh:
> Ehm I'm not sure I understand. The labels should be at the top of the circle. That's the idea. I am interested in whether that works. If you can't show it, then we'll have to wait until it's merged so I can try it.

Hi Rob 👋

Thanks for the clarification. I now fully understand that the goal is to place the labels above the circles — not inside or centered within them.

I'll make the necessary changes shortly and share updated screenshots within a few hours so you can review. Appreciate your patience!

Let me know if there’s anything else you’d like me to address while I’m at it.

PR #621, opened 2025-07-26
Fixed ticket #619 - Changes in Zoomable -Circles (Parth)

Pa04rth:
@robvanderveer @northdpole @paoga87 
As mentioned in the ticket #619 , here the changes made in the Zoomable Circles -- 
1) Changes the name from "Cluster" --> "OpenCRE" 
2) Increased the Zoom-in/out scale from "0.8" --> "2.4" (3 times) 
3) - Solved the resizing problem --> Cause = D3 circles don't resize with the browser window because the drawing logic within the useEffect hook is not re-running when the window dimensions change --> Solution = Moved size  outside the useEffect and added it to dependency 
4) Enabled the labels and made then render at the top of the circle

Demonstrating video - 

https://github.com/user-attachments/assets/75833037-6fa4-4517-a545-1469ae338e14

Extra changes discussed  After this -- 
1)- Removed the child count 
2)- Removed the Prefix "CRE" and also the number 
3)- To show full label , i removed truncate 
4)- Added down-dash and underline
Ex- 
image

### DIFF
--- a/application/frontend/src/pages/Explorer/visuals/circles/circles.scss
+++ b/application/frontend/src/pages/Explorer/visuals/circles/circles.scss
@@ -1,32 +1,61 @@
-
 .node {
-    cursor: pointer;
+  cursor: pointer;
 }
 
 .node:hover {
-    stroke: #000;
-    stroke-width: 1.5px;
+  stroke: #000;
+  stroke-width: 1.5px;
 }
 
 .node--leaf {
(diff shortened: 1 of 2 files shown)


northdpole (approved):

Pa04rth:
@northdpole I have solved all the issues

northdpole (approved):

PR #633, opened 2025-10-12
adjust the chatbot prompt to notify users that the reply isn't based …

northdpole:
…on open cre


northdpole:
Thanks @northdpole for the contribution and for putting time into OpenCRE — please keep contributing when you can.

**Closing:** Superseded by the current chat prompt / retrieval contract (`chat_prompt.j2` + retrieved-knowledge rules), not this vertex-client-only change.

PR #641, opened 2025-11-16
fix: Resolve docker-dev build failures for arm64 and macOS

### LINKED ISSUE #640: [Bug]: 'make docker-dev' fails on arm64 (Apple M1/M2) and requires setup modifications
### Environment:
1. **os:** macOS (Apple Silicon M1/M2/M3)
2. **Architecture:** arm64
3. **Setup Method:** `make docker-dev `

### Problem 1: Puppeteer/Chromium Build Failure on `arm64`
The first failure happens during the `node:lts` build stage when `yarn install` is run.

**Actual Behavior:** The build fails during yarn install with the error The chromium binary is not available for arm64.
The build fails when running `yarn install`. The root cause appears to be the `puppeteer` package. Its install script tries to download a pre-built binary of the Chromium browser. However, since the Docker build is running on an `arm64` architecture (I'm on an Apple M1/M2/M3 Mac), the script fails because it cannot find a pre-built **`arm64` Linux binary** for Chromium.

**Logs:** 
Image
Image

**Proposed FIx:** Install `chromium` via `apt` before running the make command.

### Problem 2: Missing Python Virtual Environment `(venv)`
After fixing Problem 1, the build fails again in the second `python:3.11.0` stage when installing Python dependencies.

**Actual Behavior:** The build fails when running the `make install-deps-python `command. The Makefile script (at line 59) checks for the existence of a `./venv` directory using `[ -d "./venv" ]`. This check fails because the virtual environment has not been created yet, which stops the build.

**Logs:** 
Image

Image

**Proposed FIx:** Explicitly create the virtual environment by running `python3 -m venv venv` after pip install `virtualenv` and before the `make install-deps-python` command.

### Problem 3: Port Conflict on `make docker-dev-run`
After successfully fixing both build errors, running the container fails due to a port conflict.

**Actual Behavior:** The `make docker-dev-run` command fails with a `bind: address already in use` error for port 5000. On macOS, this port is commonly occupied by the AirPlay Receiver (`ControlCe`) system service, which makes it impossible to run the container.

**Logs:**
Image

**Proposed Fix:** Change the host port mapping in the Makefile from `5000 to 5001` (or another free port) to avoid this common conflict.

3Sangeetha3:
Fixes #640

This PR addresses multiple issues that were causing the `make docker-dev` command to fail for users on `arm64` (Apple Silicon) machines.

### 1. Fixes Puppeteer/Chromium Build Failure
* **Problem:** The build failed during `yarn install` because `puppeteer` could not find a pre-built `arm64` Linux binary for Chromium.
* **Solution:** This PR adds a `RUN apt-get update && apt-get install -y chromium` step to the `Dockerfile-dev` to install the required binary before `yarn install` is run.

### 2. Fixes Missing Python 'venv' Error
* **Problem:** The build failed a second time because the `make install-deps-python` command checks for a `./venv` directory that hadn't been created.
* **Solution:** Added `python3 -m venv venv` to the Dockerfile to create the virtual environment before the `make` command is called.

### 3. Resolves macOS Port Conflict
* **Problem:** Running the container with `make docker-dev-run` failed with an "address already in use" error, as port 5000 is often used by the macOS AirPlay Receiver.
* **Solution:** Updated the `docker-dev-run` target in the `Makefile` to map to port `5001` on the host (`-p 5001:5000`) to avoid this common conflict.

With these changes, users on `arm64`/macOS can successfully build and run the development container.


Makefile
```diff
@@ -83,7 +83,7 @@ docker-prod:
 	docker build -f Dockerfile -t opencre:$(shell git rev-parse HEAD) .
 
 docker-dev-run:
-	 docker run -it -p 5000:5000 opencre-dev:$(shell git rev-parse HEAD)
+	 docker run -it -p 5001:5000 opencre-dev:$(shell git rev-parse HEAD)
```
northdpole: Why's this necessary?

northdpole (changes requested):
Thanks for the contribution! Just a couple comments and this is ready

northdpole (approved):

PR #679, opened 2025-12-28
Fix: Makefile targets fail when project path contains spaces

### LINKED ISSUE #697: Fix: Some Makefile targets fail when project path contains spaces
## Description

Many Makefile targets in OpenCRE fail when the project directory path contains spaces (e.g. `~/git folder/OpenCRE`).  
This happens because unquoted shell paths are expanded incorrectly by the shell.

Commands such as `make e2e`, `make test`, `make import-neo4j` rely on `FLASK_APP` and other path-based variables, which are set without proper quoting.

---

## Steps to Reproduce

1. Clone the repository into a directory with spaces in its name:
   ```sh
   git clone https://github.com/OWASP/OpenCRE.git "git folder/OpenCRE"
   cd git\ folder/OpenCRE
2. Create and activate a virtual environment inside the project directory:
   ```sh
   python3 -m venv venv
   source venv/bin/activate
3. Run any of the following:
   ```sh
   make test
   make e2e
   make migrate-upgrade

---

## Observed Behavior

-  Shell fails to correctly resolve paths containing spaces.
-  Error occurs,
    <img width="535" height="38" alt="Image" src="https://github.com/user-attachments/assets/66008919-645d-41a7-a815-a0848986088b" />
     ```sh
       /bin/sh: line 0: export: `folder/OpenCRE/cre.py': not a valid identifier
       make: *** [test] Error 1

-  FLASK_APP points to an invalid path when expanded.
---

## Expected Behavior

- Makefile targets should work regardless of whether the project path contains spaces.
- Shell commands should correctly resolve paths.

---
## References

#### https://mywiki.wooledge.org/Quotes
---
## Suggested fix

Quote path-based variables when exporting them in Makefile targets, for example:
 ```sh
  export FLASK_APP="$(CURDIR)/cre.py"

shreyakash24:
Fixes #697 
## Changes made
Quoted path-based variables when exporting them in Makefile targets, for example:
 ```sh
  export FLASK_APP="$(CURDIR)/cre.py"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ dev-flask-docker:
 e2e:
 	yarn build
 	[ -d "./venv" ] && . ./venv/bin/activate &&\
-	export FLASK_APP=$(CURDIR)/cre.py &&\
+	export FLASK_APP="$(CURDIR)/cre.py" &&\
 	export FLASK_CONFIG=development &&\
 	export INSECURE_REQUESTS=1 &&\
 	flask run &
@@ -52,7 +52,7 @@ e2e:
 
 test:
 	[ -d "./venv" ] && . ./venv/bin/activate &&\
-	export FLASK_APP=$(CURDIR)/cre.py &&\
+	export FLASK_APP="$(CURDIR)/cre.py" &&\
 	flask routes && flask test
 
 cover:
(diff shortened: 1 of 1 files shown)


shreyakash24:
> heh, hadn't thought about that, good catch!

Thanks! I ran into it while testing things locally.

Issue #584, opened 2025-02-22
MyOpenCRE: create frontend


northdpole:
MyOpenCRE is a year old at this point. The API seems to have reasonable stability and to work as expected.

Time to make a frontend for it so that interested parties can use it without curl.

Acceptance criteria:
Make a set of new pages and buttons linked to by the main page that allow users to:
* download a CSV with all the CREs currently in the DB
* upload a CSV with one or more standards mapped to CREs and allow the application to import them.

Ad-tech1009:
Interested in taking this up
I'd like to work on this issue and build the frontend for downloading and uploading CSVs. Before I start, could you confirm the preferred tech stack for the frontend? Should I align with any existing framework (React, Vue, or plain HTML/JS), or do you have specific requirements?

Let me know, and I'll proceed accordingly. You can assign this issue to me.

northdpole:
You should follow the existing tech, currently react and typescript, it should be kinda seamless with the existing frontend:-)

Reach out to slack If you need help, happy to add details

Ad-tech1009:
@northdpole Thanks for confirming! I'll proceed with implementing the frontend in React and TypeScript, ensuring it integrates seamlessly with the existing setup. Please assign the issue to me, and I'll get started.

abhayofc:
@Ad-tech1009 are you working on this?

robvanderveer:
@northdpole ?

PRAteek-singHWY:
Hi @northdpole and @robvanderveer,

I’ve been going through the MyOpenCRE flow and this issue aligns very well with the current frontend architecture and user journey.

I noticed the issue has been quiet for a while, so before starting anything, I wanted to check:
- Is this still a priority for the MyOpenCRE milestone?
- Are there any backend/API changes since the issue was opened that the frontend should be aware of?

If it’s still open, I’d be happy to pick this up and start with a small, incremental PR (e.g., CSV download + template flow) before moving to upload + notifications.

Happy to coordinate on Slack as well if that’s easier.

PR #682, opened 2025-12-28
feat(myopencre): improve CSV import validation (export-compatible)

PRAteek-singHWY:
⚠️ **This PR depends on:**
- #664  
Please review this PR after the above PR is merged.

---

### Note for reviewers

The intended change in this PR is limited to  
`application/web/web_main.py` (CSV import validation logic).

Any additional file diffs shown by GitHub are inherited from branch history / stacking and are **not** part of the validation change itself.

Please let me know if you’d prefer this PR to be rebased or split into a narrower diff.

---

### Summary

Closes #584 (partial)

This PR improves server-side validation for the MyOpenCRE CSV import endpoint, aligning it closely with the CSV format produced by the CRE catalogue export flow.

The intent is **not** to introduce strict or surprising validation rules, but to ensure that:

- every CSV produced by the exporter can be safely imported
- malformed or ambiguous input is rejected early with clear, structured errors
- exporter artifacts (padding rows, empty rows) do not cause failures

---

#### File-level validation
- Rejects missing, empty, non-CSV, or non–UTF-8 uploads
- Returns consistent, structured error responses

#### Schema / header validation
- Requires at least one `CRE*` column
- Requires `standard|name` and `standard|id`
- Rejects rows with more columns than the header (misaligned CSVs)

#### Row-level validation (export-compatible)
- Completely empty rows are ignored (exported templates include padding rows)
- Rows without any CRE values are ignored
- CRE entries are validated only when present (`<CRE-ID>|<Name>`)
- Malformed CRE entries return row-specific validation errors

#### No-op import guard
- If a file contains no importable rows after filtering, the request returns success with no changes
- This avoids confusing partial imports or unnecessary failures

---

### What is intentionally ignored (by design)

- Empty rows
- Padding rows from exported templates
- Rows without CRE mappings
- Extra unused columns (as long as the CSV structure itself is valid)

This mirrors how production importers typically behave and allows exported files to be round-tripped without manual cleanup.

---

### What is out of scope for this PR

- Frontend error presentation / formatting
- UX changes around errors

These will be handled in follow-up PRs to keep this change focused and reviewable.

---

### Why this approach

The importer now accepts everything the exporter produces, ignores exporter padding rows, and enforces validation only where semantic meaning exists.

This keeps the import process resilient while still preventing invalid data from entering the system.

---

### Dependency note

This PR is stacked on top of:

`feat(myopencre): add CSV upload UI and wire to existing import endpoint (#664)`

It should be reviewed and merged after that PR.

---

### Feedback welcome

If any validation behavior is too permissive or too strict, or if there are exporter edge cases I may have missed, I’d really appreciate guidance and am happy to adjust.

### DIFF
--- a/application/tests/gap_analysis_db_test.py
+++ b/application/tests/gap_analysis_db_test.py
@@ -35,8 +35,11 @@ class TestGapAnalysisPruning(unittest.TestCase):
 
         self.mock_cypher.side_effect = cypher_side_effect
 
-        # Call the function
-        db.NEO_DB.gap_analysis("StandardA", "StandardB")
+        # Call the function with tiered pruning enabled
+        with patch(
+            "application.config.Config.GAP_ANALYSIS_OPTIMIZED", True, create=True
+        ):
+            db.NEO_DB.gap_analysis("StandardA", "StandardB")
 
         # ASSERTION:
         # We expect cypher_query to be called.
--- a/application/tests/web_main_test.py
+++ b/application/tests/web_main_test.py
@@ -905,24 +905,12 @@ class TestMain(unittest.TestCase):
                 buffered=True,
                 content_type="multipart/form-data",
             )
+            print(f"\nSTATUS CODE: {response.status_code}, DATA: {response.data}")
             self.assertEqual(200, response.status_code)
-            self.assertEqual(
-                {
-                    "status": "success",
-                    "new_cres": [
-                        "000-001",
-                        "222-222",
-                        "333-333",
-                        "444-444",
-                        "555-555",
-                        "666-666",
-                        "777-777",
-                        "888-888",
-                    ],
-                    "new_standards": 5,
-                },
-                json.loads(response.data),
-            )
+            data = json.loads(response.data)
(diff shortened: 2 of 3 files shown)


application/web/web_main.py
```diff
@@ -784,25 +784,214 @@ def import_from_cre_csv() -> Any:
...
         False if not request.args.get("calculate_gap_analysis") else True
     )
 
+    # ------------------------
+    # File-level validation
+    # ------------------------
+
     if file is None:
-        abort(400, "No file provided")
+        return (
```
northdpole: Excellent work! thanks for adding the validation.
One thing: this very helpful validation is a bit confusing and out of place in the "controllers" file.
Ideally this file only handles requests/responses. Why don't we move this validation to the spreadsheet parsers file in the appropriate function?

This way not only we  get a bit clearer code but also we get a start on a spreadsheet validation utility we can eventually make into its own module.
For now just moving this validation logic into the appropriate function in the spreadsheet handlers module is enough :+1:

northdpole (approved):

PR #696, opened 2026-01-03
Fix: improved caching efficiency to increase rebuild speed

### LINKED ISSUE #695: Improve Dockerfile-dev layer caching to reduce rebuild time
## Description

Currently, frontend dependencies are reinstalled on every Docker build, even when there are no changes to `package.json` or `yarn.lock`. This causes unnecessary rebuild time, especially during local development.
When working on OpenCRE locally, developers often rebuild the dev Docker image after making small changes (e.g. Python code, configs, or documentation).  
Reinstalling frontend dependencies on every build significantly slows down this workflow.

---

1. Build the development Docker image:
   ```sh
   docker build -f Dockerfile-dev -t opencre-dev:test .
2. Make a non-dependency change (for example, modify a Python or JS file).
3. Rebuild the image:
   ```sh
   docker build -f Dockerfile-dev -t opencre-dev:test .

---

## Observed Behavior

-  The yarn install step runs on every build.
-  Frontend dependency installation significantly increases rebuild time.
---

## Expected Behavior

- Rebuilds after non-dependency changes should skip frontend dependency installation.
- Faster and more efficient local development builds.

---
## References
#### https://docs.docker.com/build/cache/optimize/
---

shreyakash24:
Fixes #695 
## Changes made

The Dockerfile has been restructured to copy dependency files first and run `yarn install` before copying the rest of the source code, allowing Docker to reuse cached layers when dependencies have not changed thereby reducing rebuild time from ~550 seconds to ~300 seconds.

Before change:-
Image

After change:-
Image

- Copied package.json and yarn.lock before the rest of the source code.
- Executed yarn install immediately after copying dependency files.
- Preserved existing build logic and behavior.

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -1,9 +1,12 @@
 FROM node:lts AS build
 
 WORKDIR /code
-COPY . /code
-
 RUN apt-get update && apt-get install -y chromium
+COPY package.json yarn.lock ./
+RUN yarn install
+
+COPY . .
+
 RUN make install-deps-typescript && make install-typescript && make frontend
 
 FROM python:3.11.0 AS run


northdpole (approved):
thank you!

PR #705, opened 2026-01-10
Display model name used

robvanderveer:
Code was fully Cursor generated and reviewd by me

### DIFF
--- a/application/prompt_client/prompt_client.py
+++ b/application/prompt_client/prompt_client.py
@@ -499,4 +499,9 @@ class PromptHandler:
         table = [closest_object]
         result = f"Answer: {answer}"
         model_name = self.ai_client.get_model_name() if self.ai_client else "unknown"
-        return {"response": result, "table": table, "accurate": accurate, "model_name": model_name}
+        return {
+            "response": result,
+            "table": table,
+            "accurate": accurate,
+            "model_name": model_name,
+        }


PRAteek-singHWY:
Hey @robvanderveer @northspyroso 👋
I recently ran into the same recurring PYTHON_BLACK lint failures and opened PR #707 to standardize Python formatting using pre-commit.
I thought it might be helpful context here — please let me know if this looks like the right long-term approach or if you’d suggest adjustments.
Thanks as well — hitting this error after solid PR work made it clear this was worth fixing 🙂

robvanderveer:
Thanks @PRAteek-singHWY, this is indeed a nuisance it seems.

PRAteek-singHWY:
> Thanks @PRAteek-singHWY, this is indeed a nuisance it seems.

Thanks for confirming, @robvanderveer  — appreciate it.

northdpole:
> Hey @robvanderveer @northdpole 👋 I recently ran into the same recurring PYTHON_BLACK lint failures and opened PR #707 to standardize Python formatting using pre-commit. I thought it might be helpful context here — please let me know if this looks like the right long-term approach or if you’d suggest adjustments. Thanks as well — hitting this error after solid PR work made it clear this was worth fixing 🙂

I'\d love for this to be pre-commit hooks, using it in the github ci can be a pain since in rare cases linting can produce non-functional code

We can easily make this pre-commit hooks using a pre-commit framework however, that'd be great

PR #714, opened 2026-01-13
Fix: added variable port

### LINKED ISSUE #713: Make Flask development port configurable to prevent port conflicts
## Description

Currently, the development server port is hardcoded to 5000 across multiple Makefile targets. This causes frequent port conflicts during local development—especially on macOS, where port 5000 is often already occupied by system services (e.g., AirPlay Receiver or other background processes).

---

## Steps to Reproduce

1. Start the development server.
    ```sh
     make dev-flask
2. If port 5000 is already in use, Flask fails to start with:
    ```wasm
    Address already in use

---

## Observed Behavior

- Port 5000 is hardcoded in multiple Makefile targets.
- Developers cannot easily override the port.
- Conflicts are common when port 5000 is already in use.

<img width="872" height="62" alt="Image" src="https://github.com/user-attachments/assets/8e79f727-707f-4d15-98cd-037c245b70e8" />

- This slows down onboarding and local development.
---

## Expected Behavior

- The development port should be configurable.
- Default behavior should remain unchanged if the variable is not set.
- Developers should be able to run:
   ```sh
   make dev-flask PORT=8000

---
## References
[Port conflicts](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://flask.palletsprojects.com/en/stable/server/&ved=2ahUKEwj87qunv42SAxXPzzgGHbGkAdsQFnoECEIQAQ&sqi=2&usg=AOvVaw1KzVhsTWIeA6ixbtapJykz)

shreyakash24:
Fixes #713

This PR introduces a configurable PORT variable with a default value of 5000, allowing developers to easily override it without modifying source files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,13 @@ start-worker:
 upstream-sync:
 	. ./venv/bin/activate && python cre.py --upstream_sync
 
+PORT?=5000
+
 dev-flask:
-	. ./venv/bin/activate && INSECURE_REQUESTS=1 FLASK_APP=`pwd`/cre.py  FLASK_CONFIG=development flask run
+	. ./venv/bin/activate && INSECURE_REQUESTS=1 FLASK_APP=`pwd`/cre.py  FLASK_CONFIG=development flask run --port $(PORT)
 
 dev-flask-docker:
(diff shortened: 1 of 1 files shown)


northdpole:
can you please sign your commits?

PR #720, opened 2026-01-16
Enable OpenCRE as selectable standard in map-analysis

### LINKED ISSUE #469: Add OpenCRE itself to mapanalysis
It would be good to have opencre be one of the standards in the map analyis. it can be helpful to have a opencre-samm overview for example.

ashutosh-engineer:
This PR enables OpenCRE to be selected on either the left or right side in the map-analysis view.

- UI-only change
- No data or ontology changes
- Uses existing mappings

Fixes #469

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,7 +46,6 @@ v/
 
 ### Frontend 
 node_modules/
-frontend/www/bundle.js
 
 .DS_Store
 yarn-error.log


ashutosh-engineer:
@robvanderveer @northdpole Before starting backend work, just to confirm:
Is OpenCRE already present in the database and graph as a standard that can be queried for map-analysis, or should this work strictly reuse existing CRE relationships without introducing new data?

ashutosh-engineer:
> Dear @ashutosh-engineer , with all due respect: based on that question I have to conclude that your understanding of the OpenCRE data structure fundementals is currently insufficient to make a reliable change. The essence of OpenCRE is that a standard links to CREs and because other standards also link to those CREs, they link to each other. In other words, OpenCRE forms an intrinsic part of the graph. I strongly advice not to go ahead and make changes before this understanding is solid.

Thank you for the clarification — that helps, and I appreciate you calling this out.

You’re absolutely right: OpenCRE is not a standard alongside others, but the central part of the graph that enables standards to relate to each other via shared CREs. My question came from trying to be careful about scope, not from an intent to treat OpenCRE as an external or optional data source.

I’ll pause any backend-related work for now and spend more time solidifying my understanding of the OpenCRE data model and graph fundamentals before proposing and modifying my PR.

Thanks again for the guidance.

ashutosh-engineer:
> Dear @ashutosh-engineer , with all due respect: based on that question I have to conclude that your understanding of the OpenCRE data structure fundementals is currently insufficient to make a reliable change. The essence of OpenCRE is that a standard links to CREs and because other standards also link to those CREs, they link to each other. In other words, OpenCRE forms an intrinsic part of the graph. I strongly advice not to go ahead and make changes before this understanding is solid.

Thanks for the earlier clarification — I spent some time going through the codebase and the map-analysis logic to make sure I fully understand the model.

My understanding now is:
1. OpenCRE is not a standard document, but the CRE layer of the graph itself.
2.This is a read-only graph traversal concern, not a data modeling or ontology change.
3.Existing map-analysis already works as: Standard → CRE → Standard, where CREs are currently implicit.
4.Supporting OpenCRE in map-analysis is about exposing those existing CRE nodes by adjusting query patterns, not adding data or creating new nodes.

Concretely, this means:
1.Matching all `NeoCRE` nodes (without filtering by a “standard name”)
2. Allowing CRE nodes as valid endpoints in the traversal
3.Keeping everything read-only, with no schema or mapping changes

If this understanding aligns with your expectations, please let me know if it’s okay for me to proceed with  backend change following this approach.

Thanks for the guidance, Sir

ashutosh-engineer:
@robvanderveer The limited set of standards was intentional and only meant to validate routing and traversal behavior locally.

PR #722, opened 2026-01-17
Tags pagination

### LINKED ISSUE #205: make pagination also for tag results
## Issue
get by tag is not paginated, we don't have a lot of tags yet so for now its performant, but paginate it so that it's faster with more data

atif09:
## Summary
This PR adds pagination support to the /rest/v1/tags endpoint to handle large result sets efficiently.

## Details
- Added get_by_tags_with_pagination() method to database layer
- Updated /rest/v1/tags endpoint to accept page and items_per_page query parameters
- Returns total_pages and page in JSON response
- Maintains backward compatibility (works without pagination parameters)
- Added comprehensive tests for pagination logic

##  Why this change
The /rest/v1/tags endpoint had a TODO comment indicating pagination was needed. Without pagination, queries with many tags could return large result sets, causing performance issues and poor UX. This change enables efficient data retrieval with client-controlled page sizes.

## Scope
- Backward compatible (pagination parameters are optional)
- Fixes #205

### DIFF
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -884,6 +884,34 @@ class Node_collection:
                 )
         return documents
 
+    def get_by_tags_with_pagination(
+        self, tags: List[str], page: int = 1, items_per_page: int = 20
+    ) -> Tuple[Optional[int], List[cre_defs.Document]]:
+
+        if not tags:
+            return 0, []
+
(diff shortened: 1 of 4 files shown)


northdpole:
Thank you! Tags pagination is covered by the broader #760 (tags + text_search). Closing as superseded.

Issue #523, opened 2024-07-06
fix e2e tests


northdpole:
### **What is the issue?**
e2e tests are failing, we should fix them

iagoscm:
hey @northdpole, where exactly are the e2e tests?

northdpole:
i think since we started migrating to cypress lets finish it

northdpole:
Closing as superseded by the Cypress migration track.

Please continue under #967 (and active PR #984). Older Cypress attempts (#729 closed, #774 conflicting) are folded into that consolidation.

PR #732, opened 2026-01-31
chores: integrate Cypress tests into Makefile

ghost:
Related issue:#523 
**Changes Made **
- Added new Cypress targets:
- cypress-open
- cypress-run
- install-cypress 
**Updated**
- e2e target to use yarn cypress run instead of playwright 
- install-python to remove playwright install and add yarn install 
- added || true to killall coammands to prevent errors


tanvikapandey7:
I'll work on it .

PR #743, opened 2026-02-12
added workflow for issue link check

### LINKED ISSUE #742: no rule for issue PR linking
current workflows doesn't have any workflow which informs the rule of a PR to must point to an issue.

manshusainishab:
Closes #742

I read the recent PR made by @northdpole in which we were having a problem of having PR without a issue which is solved by this PR.

### DIFF
--- a/.github/workflows/require-issue.yml
+++ b/.github/workflows/require-issue.yml
@@ -1,0 +1,33 @@
+name: Auto Close PR Without Linked Issue
+
+on:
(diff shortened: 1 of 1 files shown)


northdpole:
thanks for the pr, unfortunately we can't merge this as the original pr where i referenced this issue is supposed to fix it, we'll merge it eventually.

looking forward to you addressing another issue

PR #771, opened 2026-02-28
Fix(searchBar): icon layout fix

### LINKED ISSUE #770: UI break in recent merge
if we open the website in Safari we can see a black div rendering top left corner and on both Safari and chrome the Search input in navbar where icon is overlapping input box and text seems to be tightly packed over input box

Image of main site on SAFARI

Image

in local run I found the same input box totally missing and a different UI

Image

manshusainishab:
fixes #770

Screenshot 2026-03-01 at 1 41 49 AM

icon layout is fixed no overlapping with the input box.

### DIFF
--- a/application/frontend/src/pages/Search/components/SearchBar.scss
+++ b/application/frontend/src/pages/Search/components/SearchBar.scss
@@ -16,8 +16,8 @@
   }
 
   .search-icon {
-    position: absolute;
-    left: 0.75rem;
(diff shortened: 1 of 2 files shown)


manshusainishab:
@northdpole fixed the icon but not sure about the position of "Search OpenCRE" if you please clarify I can make the changes I have attached the current state of the page.

PR #788, opened 2026-03-05
fix(explorer): restore readable text color regression from #784

PRAteek-singHWY:
## Summary
Fixes a text-contrast regression introduced by #784 on `/explorer`, where heading/body text became near-white on a light background.

## What changed
- Set default Explorer text color back to black.
- Set heading/paragraph/label/menu-title color back to black.

## Before
image

## After
image

## Verification

- Opened `/explorer` and confirmed text is readable without selecting/highlighting.


PRAteek-singHWY:
> @PRAteek-singHWY if I am not wrong the original theme of the background was not white it is introduced in PR #768 and now in this PR we are again struggling with readability issue because of theme changes. Is this theme change planned for the website? If not I propose to get theme from system preference and our website should be dynamic for both dark and light.

Hi @manshusainishab 
No new background color was introduced in PR #768. That PR mainly restored the body background to white, which had been changed earlier in PR #767.
In PR #767, the background was modified from white to #010817.

That PR was raised to resolve the issue where a white background appeared during overscrolling. PR #768 addressed the overscrolling behavior itself and restored the default body background to white. You can also check the video attached in PR #768 to see what was breaking earlier after PR #767
Regarding the readability issue on the explorer, it is already being handled in PR #788.

manshusainishab:
Thanks for the clarification @PRAteek-singHWY I thought the original colour was not white because when I was fixing the readability issue I applied the text with white colour because background was dark at that time on main branch

PR #790, opened 2026-03-06
feat: list supported standards and documents

### LINKED ISSUE #383: List supported standards and documents
## Issue
[From Sam Stepanyan]
It would be convenient to know which standards and in general documents are supported by opencre
this helps both with chatbot queries but also general knowledge of what OpenCRE supports

Implementation wise this is a simple query to select unique names from nodes and a frontend page to present this somehow

PRAteek-singHWY:
## Summary
- add a new `/rest/v1/supported_documents` API endpoint that returns unique node names grouped by document type
- add a new frontend page at `/supported_documents` to display supported sources in grouped cards
- make each source name clickable to existing OpenCRE routes (`/node/<type>/<name>`) for drill-down
- add a `Sources` navigation link in desktop and mobile headers
- document the endpoint in OpenAPI and add backend tests

## Why
Issue #383 asks for a clear list of standards/documents supported by OpenCRE so users and chatbot workflows can discover available sources quickly.

## Validation
- ./venv/bin/python -m unittest application.tests.db_test.TestDB.test_supported_documents application.tests.web_main_test.TestMain.test_supported_documents_from_db

- npm run build

##Screenshot

image

Closes #383

### DIFF
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -1811,7 +1811,9 @@ class Node_collection:
 
         return {
             ntype: sorted(list(names), key=lambda value: value.lower())
-            for ntype, names in sorted(documents_by_type.items(), key=lambda item: item[0].lower())
+            for ntype, names in sorted(
+                documents_by_type.items(), key=lambda item: item[0].lower()
+            )
         }
 
     def text_search(self, text: str) -> List[Optional[cre_defs.Document]]:
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -1798,6 +1798,22 @@ class Node_collection:
         )
         return list(set([s[0] for s in standards]))
 
+    def supported_documents(self) -> Dict[str, List[str]]:
+        documents_by_type: Dict[str, set[str]] = {}
+        for ntype, name in (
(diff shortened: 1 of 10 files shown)


northdpole:
**Maintainer note / product question**

On `main` today the frontend does **not** call any `supported_documents` endpoint. `/rest/v1/standards` already lists standard names.

This PR would **add** a new Sources-style page plus `/supported_documents` grouping by doctype (issue #383).

Unless we still want that Sources UI from #383, this PR has no production consumer and we should close it. Leave #383 open only if product still wants that UI.

northdpole:
Thanks @PRAteek-singHWY for the contribution and for putting time into OpenCRE — please keep contributing when you can.

**Closing:** No production consumer right now — on `main` the frontend does not use a `supported_documents` API/page, and `/rest/v1/standards` already lists standard names. We are not pursuing the Sources-style UI from this PR at the moment.

Issue #383 can stay open if product still wants a Sources page later; otherwise it can be closed separately.

The work was appreciated. Fresh PRs against current `main` are always welcome.

PR #791, opened 2026-03-06
fix(frontend): replace 'CREs' with 'Requirements' in user-facing text

### LINKED ISSUE #453: Use 'Requirement' instead of CRE
New visitors don't know what a CRE is. Besides it's common requirement ENUMERATION. So it should in fact read CR. Let's call it Requirement everywhere. To start with at the places where we say "Which contains CREs" and "Which is related to CREs"

rashmivemuri:
Fixes #453

## Problem
New visitors to OpenCRE don't know what a "CRE" is, making the UI 
confusing for first-time users.

## Changes
Updated user-facing display strings in `application/frontend/src/const.ts`:
- "is part of CREs" → "is part of Requirements"
- "is linked from CREs" → "is linked from Requirements"  
- "contains CREs" → "contains Requirements"
- "is related to CREs" → "is related to Requirements"

Note: `bundle.js` is auto-generated and will update automatically on rebuild.
Please let me know if changes to `docs/implementation.md` or any other 
files are also required and I'll update the PR.

### DIFF
--- a/application/frontend/src/const.ts
+++ b/application/frontend/src/const.ts
@@ -10,10 +10,10 @@ export const TYPE_AUTOLINKED_TO = 'Automatically linked to';
 export const DOCUMENT_TYPE_NAMES = {
   [TYPE_AUTOLINKED_TO]: ' has been automatically mapped to',
   [TYPE_LINKED_TO]: ' is linked to sources',
-  [TYPE_IS_PART_OF]: ' is part of CREs',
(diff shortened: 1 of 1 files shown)


northdpole:
Closing as stale and conflicting.

This was a 4-line frontend string swap (CREs → Requirements in `const.ts`, #453). The branch has been DIRTY since March with no viable rebase path. We still want friendlier wording, but it should land in a fresh PR against current UI copy.

Thank you @rashmivemuri.

PR #822, opened 2026-03-22
fix: correct tag handling in dbNodeFromCode

### LINKED ISSUE #821: Incorrect tag handling in dbNodeFromCode leads to empty tags
### Summary

While looking at dbNodeFromCode() in application/database/db.py, I noticed that tags are joined incorrectly when code.tags is non-empty.

It currently does ",".join(tags) where tags is still an empty string, instead of joining code.tags.

### Expected behavior

Code nodes should store their tags as a comma-separated string (similar to how dbNodeFromStandard() works).

### Actual behavior

Tags for Code nodes are always stored as empty, even when code.tags contains values.

### Suggested fix

**Replace the join with:**

tags = ",".join(code.tags)

### Impact

Tag metadata for Code nodes is lost
Features relying on tags (filtering, exports, etc.) may behave incorrectly

I came across this while exploring the codebase during my GSoC preparation — happy to work on a fix if this looks good.

akashrajeev:
Fixes #821 

## Summary

This PR fixes incorrect tag handling in `dbNodeFromCode` in `application/database/db.py`.

Tags were being joined from an empty variable instead of `code.tags`, causing them to always be stored as empty even when values were present.

## Changes

* Updated tag join logic to use `code.tags`

## Impact

* Fixes loss of tag metadata for Code nodes
* Improves correctness of stored data and related features

## Testing

* Verified locally that tags are now stored correctly
* Confirmed no regressions in existing functionality

### DIFF
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -2060,7 +2060,7 @@ def dbNodeFromCode(code: cre_defs.Node) -> Node:
     code = cast(cre_defs.Code, code)
     tags = ""
     if code.tags:
-        tags = ",".join(tags)
+        tags = ",".join(code.tags)
     return Node(
         name=code.name,
         ntype=code.doctype.value,


Pa04rth:
@akashrajeev  Sync and sign your PR

PR #717, opened 2026-01-15
feat(frontend): Refine Gap Analysis link strength categorization (Weak threshold 20 -> 7)

PRAteek-singHWY:
> [!NOTE]
> **Stacked PR**: This PR is built on top of **PR #716** (Backend Optimization). Please merge #716 first.

## Summary
This PR adjusts the categorization logic for Gap Analysis links in the frontend. It lowers the score threshold for "Weak" links from `20` to `7` to better reflect the true quality of connections, as requested by the maintainers.

## Motivation
The maintainer noted:
> "By the way, I notice we don't seem to have weak links anymore: should we change our categorization and make 3-6 average and 7+ weak?"

Previously, a "Weak" link was defined as having a penalty score of `20` or higher. This score is practically unreachable in normal operation (requires ~10 hops of "related" links), causing truly weak connections to be mislabeled as "Average" or "Strong". This change aligns the UI with the maintainer's request and the user's intuitive understanding of link strength.

## Changes
*   **[application/frontend/src/pages/GapAnalysis/GapAnalysis.tsx](cci:7://file:///Users/prateeksingh/OpenCRE/application/frontend/src/pages/GapAnalysis/GapAnalysis.tsx:0:0-0:0)**:
    *   Updated [GetStrength](cci:1://file:///Users/prateeksingh/OpenCRE/application/frontend/src/pages/GapAnalysis/GapAnalysis.tsx:43:0-48:2) function: Changed condition for 'Weak' from `>= 20` to `>= 7`.
    *   Updated [GetStrengthColor](cci:1://file:///Users/prateeksingh/OpenCRE/application/frontend/src/pages/GapAnalysis/GapAnalysis.tsx:50:0-55:2) function: Changed condition for Red color from `>= 20` to `>= 7`.
    *   Updated the UI Popup Legend to explicitly state that "Weak" links start at score 7.

## Related Issue
*   Fixes #506 
*   Part 2 of "Optimize Gap Analysis Pruning".

## Verification
1.  Navigate to the map analysis page.
2.  Hover over the strength indicator (Popup).
3.  Verify the legend states: `"Weakly connected likely to have small or no overlap"` corresponds to score **7**.
4.  Verify that paths with a score of 7+ appear as **Red/Weak**.

## Checklist
- [x] **Backend Pruning**: Complete (Implemented in PR #716)
- [x] **Frontend Categorization**: Refined "Weak" threshold to 7+ (This PR)

PR #827, opened 2026-03-26
Fix Session.delete() handling in gap analysis cleanup (#826)

### LINKED ISSUE #826: Fix incorrect Session.delete() result handling in gap analysis cleanup
## Summary

While looking through `delete_gapanalysis_results_for()` in `application/database/db.py`, I noticed that `Session.delete()` is being used as if it returns a result object.

However, in SQLAlchemy ORM, `session.delete()` returns `None`, so the current logging logic doesn’t behave as intended.

## Current behavior

The method does something like:

* `result = self.session.delete(r)`
* `if result: logger.info(...)`

Since `delete()` returns `None`, this condition never triggers, and the logging branch is effectively dead.

## Expected behavior

* Rows should still be deleted correctly
* Logging should reflect actual deletions using a reliable value (e.g. number of rows processed)

## Why this matters

This can be confusing for contributors reading the code, since it suggests `delete()` returns metadata like `rowcount`, which isn’t the case.

Cleaning this up would make the code easier to understand and avoid misleading patterns.

## Suggested approach

* Remove reliance on the return value of `session.delete()`
* Optionally log based on the number of matched rows (e.g. length of query result or bulk delete count)

## Scope

Small, localized fix in:
`application/database/db.py` → `delete_gapanalysis_results_for()`

I’ll go ahead and open a PR for this—happy to adjust it based on any feedback.

akashrajeev:
## Summary

This PR fixes an issue with how cached gap analysis rows are deleted in `delete_gapanalysis_results_for()`.

Previously, `Session.delete()` was used incorrectly, which caused unexpected behavior during the gap analysis cleanup process. This update corrects the logic to align with SQLAlchemy 2.x expectations and ensures proper deletion of cached results.

Note: This PR focuses only on the SQLAlchemy cleanup fix. The tiered Neo4j gap-analysis optimization work was already merged separately in #716, #717, and #748.

### application/database/db.py

* Fixed `Session.delete()` handling within `delete_gapanalysis_results_for()`
* Ensured proper cleanup of gap analysis results

### application/tests/db_test.py

* Updated tests to cover gap analysis result cleanup
* Improved coverage for associated logging

## How to Test

- [x] Run test suite (focus on db_test.py)
- [x] Ensure all tests pass and CI is green

### DIFF
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -1531,10 +1531,16 @@ class Node_collection:
             .filter(GapAnalysisResults.cache_key.like(f"%{node_name}%"))
             .all()
         )
+        deleted_count = 0
         for r in res:
-            result = self.session.delete(r)
-            if result:
-                logger.info(f"deleted {result.rowcount} objects")
+            self.session.delete(r)
+            deleted_count += 1
+        if deleted_count:
+            logger.info(
+                "deleted %s gap analysis result objects for node %s",
+                deleted_count,
+                node_name,
+            )
         self.session.commit()
         self.session.flush()
         return res
--- a/application/tests/db_test.py
+++ b/application/tests/db_test.py
@@ -2158,6 +2158,27 @@ class TestDB(unittest.TestCase):
             self.collection.standards(),
         )
 
+    def test_delete_gapanalysis_results_for_deletes_and_logs_count(self):
+        node_name = "BarStand"
+        key_1 = f"cache-{node_name}-1"
+        key_2 = f"cache-{node_name}-2"
+        unrelated_key = "cache-unrelated"
+
+        self.collection.add_gap_analysis_result(key_1, "{}")
+        self.collection.add_gap_analysis_result(key_2, "{}")
+        self.collection.add_gap_analysis_result(unrelated_key, "{}")
+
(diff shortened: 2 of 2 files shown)


northdpole:
**Review:** Correct fix — `session.delete()` returns None in SQLAlchemy 2.x; counting deletions explicitly is the right approach. Tests added.

**Blockers:** CONFLICTING on main — needs rebase. Otherwise LGTM.

northdpole:
GSoC is over — we're driving this rebase ourselves.

[#827](https://github.com/OWASP/OpenCRE/pull/827) is still CONFLICTING / not rebaseable on `main`. The same fix (plus a test that matches current empty-primary GA cache rules) is in **#1099**. We'll land that and close this as superseded.

Thanks @akashrajeev — original patch is yours.

northdpole:
Rebase is on current `main` (including the mapping-fixture loader) in #1099. Still waiting on CODEOWNERS; CI was green before this main merge.

northdpole:
Landed via #1099 (admin rebase-merge onto current main). Closing as superseded.

Thanks @akashrajeev — the Session.delete() / #826 fix is yours.

PR #849, opened 2026-03-31
fix: add pagination to /rest/v1/root_cres and cap per_page on list endpoints

### LINKED ISSUE #847: Unbounded API responses on /rest/v1/root_cres and /rest/v1/all_cres
### **What is the issue?**

Two REST endpoints can return unbounded result sets, causing high memory usage and potential timeouts in production.

`/rest/v1/root_cres` has no pagination at all. It calls `get_root_cres()` which runs a full table scan with no limit and returns every root CRE in a single response. There is no way for clients to page through results.

`/rest/v1/all_cres` already has pagination but accepts any positive integer for `per_page`. Passing `?per_page=999999` fetches the entire dataset in one request, defeating pagination entirely. There is no `MAX_PER_PAGE` cap anywhere in the codebase.

### Expected Behaviour

`/rest/v1/root_cres` should support `page` and `per_page` query parameters and return pagination metadata (`page`, `total_pages`) alongside `data`, consistent with how `/rest/v1/all_cres` already works.

`/rest/v1/all_cres` should enforce an upper bound on `per_page` so that a single request cannot retrieve the entire dataset regardless of what value is passed.

### Actual Behaviour

`/rest/v1/root_cres` returns all root CREs in a single unbounded response with no pagination support.

`/rest/v1/all_cres` accepts `?per_page=999999` and returns the entire dataset in one response.

### Steps to reproduce

```
GET /rest/v1/root_cres
# returns all root CREs, no page/total_pages in response

GET /rest/v1/all_cres?per_page=999999
# returns entire dataset in one response
```

shiwani42:
Fixes #847

### `/rest/v1/root_cres`

Added `get_root_cres_with_pagination(page, per_page)` to `db.py` using SQLAlchemy's `.paginate()`. The endpoint now accepts `page` and `per_page` query params and returns `page` and `total_pages` alongside `data` in the JSON response, consistent with `/rest/v1/all_cres`.

Format-based responses (Markdown, CSV, OSCAL) are intentional full-export flows and remain unpaginated.

### `/rest/v1/all_cres`

Added `MAX_PER_PAGE = 100` constant. `per_page` is now capped at `MAX_PER_PAGE` on both list endpoints, preventing a single request from fetching the entire dataset.

## Tests

- Added `test_get_root_cres_with_pagination` in `db_test.py`
- Updated `test_find_root_cres` in `web_main_test.py` to reflect the new response shape
- Added `test_root_cres_per_page_cap` and `test_all_cres_per_page_cap` in `web_main_test.py`

### DIFF
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -1915,6 +1915,32 @@ class Node_collection:
             result.extend(self.get_CREs(external_id=c.external_id))
         return result
 
+    def get_root_cres_with_pagination(self, page: int = 1, per_page: int = 20) -> tuple:
+        """Returns paginated root CREs (those that only have "Contains" links)"""
+        cres_page = (
+            self.session.query(CRE)
+            .filter(
+                ~CRE.id.in_(
+                    self.session.query(InternalLinks.cre).filter(
+                        InternalLinks.type == cre_defs.LinkTypes.Contains,
+                    )
+                )
+            )
+            .filter(
+                ~CRE.id.in_(
+                    self.session.query(InternalLinks.group).filter(
+                        InternalLinks.type == cre_defs.LinkTypes.PartOf,
+                    )
+                )
+            )
+            .paginate(page=page, per_page=per_page, error_out=False)
+        )
+        total_pages = cres_page.pages
+        result = []
+        for c in cres_page.items:
+            result.extend(self.get_CREs(external_id=c.external_id))
+        return result, page, total_pages
+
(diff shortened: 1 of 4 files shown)


northdpole:
Thanks @shiwani42 for the contribution and for putting time into OpenCRE — please keep contributing when you can.

**Closing:** Not needed: we only have a handful of root CREs, so paginating `/root_cres` is not valuable; `MAX_ITEMS_PER_PAGE` already exists for larger list endpoints.

PR #852, opened 2026-03-31
feat: validate critical config at startup and abort with clear messages

### LINKED ISSUE #851: Startup validation: abort early on missing critical configuration
### **What is the issue?**

The application does not validate critical configuration at startup. Missing environment variables are only discovered at runtime when a specific feature is exercised, leading to confusing errors deep in the call stack.

Three cases stand out:

**Web server:**
`GOOGLE_CLIENT_SECRET` is assigned directly as `app.secret_key` with no check. If it is not set, Flask runs with `secret_key = None`, which silently breaks session signing in production. `GOOGLE_CLIENT_ID` is similarly absent with no warning, so the OAuth flow fails only when a user attempts to log in.

**`--generate_embeddings` CLI command:**
No AI provider key check occurs before the command starts. The missing-key error is only logged inside `PromptHandler.__init__`, which is called well into the import process, leaving the user with an opaque failure.

**`--populate_neo4j_db` CLI command:**
`NEO4J_URL` defaults silently to `localhost:7687`. If that instance is not running, the failure surfaces as a connection error inside the Neo4j driver rather than a clear startup message.

### Expected Behaviour

The application should check for required configuration as early as possible and exit immediately with a human-readable error message that names the missing variable and explains how to fix it.

Optional services (Neo4j, Redis) that are not configured should produce a visible warning at startup so the operator knows which features will be degraded before any request is served.

### Actual Behaviour

- Flask starts with `secret_key = None` when `GOOGLE_CLIENT_SECRET` is not set.
- `--generate_embeddings` fails inside `PromptHandler` with a logged error but no early exit.
- `--populate_neo4j_db` fails with a Neo4j driver connection error rather than a clear config message.
- No warnings are produced for missing optional services.

### Steps to reproduce

```bash
# Start the web server without setting GOOGLE_CLIENT_SECRET
flask run
# Sessions are broken; no startup error is shown

# Run embeddings generation without an AI provider key
python cre.py --generate_embeddings
# Fails inside PromptHandler with a logged error, no early exit

# Run Neo4j population without NEO4J_URL
python cre.py --populate_neo4j_db
# Fails with a Neo4j driver connection error
```

shiwani42:
Fixes #851

## What changed

Added `application/validate_config.py` with three focused validators:

**`validate_web_config(mode)`**
Called from `create_app()` when running in web server mode. Checks that `GOOGLE_CLIENT_SECRET` and `GOOGLE_CLIENT_ID` are set and exits with a clear error if either is missing and `NO_LOGIN` is not set. Logs warnings when `NEO4J_URL` or `REDIS_URL` are absent. Skipped entirely in test/testing mode and when a custom `conf` object is provided (CLI `db_connect` path).

**`validate_embeddings_config()`**
Called in `cre_main.run()` immediately before `--generate_embeddings`. Exits if none of `OPENAI_API_KEY`, `GEMINI_API_KEY`, or `GCP_NATIVE` is configured.

**`validate_neo4j_config()`**
Called in `cre_main.run()` immediately before `--populate_neo4j_db`. Exits if `NEO4J_URL` is not set.

## Tests

13 new tests in `application/tests/validate_config_test.py` covering:
- test mode is skipped entirely
- exits when Google credentials are missing
- skips auth check when `NO_LOGIN` is set
- passes when all required credentials are present
- warns but does not exit for missing optional services
- exits for missing AI provider before embeddings generation
- exits for missing `NEO4J_URL` before Neo4j population

### DIFF
--- a/application/__init__.py
+++ b/application/__init__.py
@@ -37,6 +37,12 @@ def create_app(mode: str = "production", conf: any = None) -> Any:
         letters = string.ascii_lowercase
         app.secret_key = "".join(random.choice(letters) for i in range(20))
 
+    # conf is only provided by CLI db_connect(); skip web validation in that path.
+    if conf is None:
+        from application.validate_config import validate_web_config
+
+        validate_web_config(mode)
+
     # config[mode].init_app(app)
     sqla.init_app(app=app)
     from application.web.web_main import app as app_blueprint
--- a/application/cmd/cre_main.py
+++ b/application/cmd/cre_main.py
@@ -656,8 +656,14 @@ def run(args: argparse.Namespace) -> None:  # pragma: no cover
         BaseParser().call_importers(db_connection_str=args.cache_file)
 
     if args.generate_embeddings:
+        from application.validate_config import validate_embeddings_config
+
+        validate_embeddings_config()
         generate_embeddings(args.cache_file)
     if args.populate_neo4j_db:
+        from application.validate_config import validate_neo4j_config
+
+        validate_neo4j_config()
         populate_neo4j_db(args.cache_file)
     if args.start_worker:
         from application.worker import start_worker
--- a/application/tests/validate_config_test.py
+++ b/application/tests/validate_config_test.py
@@ -1,0 +1,150 @@
+import os
+import unittest
+from unittest.mock import patch
+
+from application.validate_config import (
+    validate_embeddings_config,
+    validate_neo4j_config,
+    validate_web_config,
+)
+
+
+class TestValidateWebConfig(unittest.TestCase):
+    def test_skips_in_test_mode(self):
+        """Should not call sys.exit in test or testing mode regardless of env vars."""
+        with patch("sys.exit") as mock_exit, patch.dict(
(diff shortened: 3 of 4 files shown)


northdpole:
**Review:** Good idea — fail-fast startup validation with clear messages. Tests cover web/embeddings/neo4j paths and NO_LOGIN bypass.

**Blockers before merge:**
- CONFLICTING on current main — needs rebase
- `validate_web_config` exits when Google OAuth missing even in local dev unless `NO_LOGIN=True` — confirm that's intended (may surprise contributors)
- Worth a quick check that Heroku prod always has credentials set (it does)

Recommend rebase + merge after GSoC crunch unless we want stricter local dev guardrails now.

shiwani42:
Sure, I'll follow up on this after GSoC

northdpole:
GSoC is over — this is still a good idea and still CONFLICTING.

Please rebase onto current `main` when you can. Remaining merge blocker from the Aug review: `validate_web_config` should not exit on missing Google OAuth in local/dev when `NO_LOGIN` is unset; fail-fast belongs in production/web mode.

Happy to review a fresh rebase.

PR #856, opened 2026-03-31
Add centralized structured logging, remove per-module basicConfig calls

### LINKED ISSUE #855: Centralize logger configuration and adopt structured logging
### **What is the issue?**

Every module in the application independently calls `logging.basicConfig()` with no arguments and hardcodes `logger.setLevel(logging.INFO)`. There are 25 such occurrences across the codebase.

This pattern has two concrete problems:

1. `logging.basicConfig()` is a no-op after the first call. Whichever module happens to be imported first claims the root logger; every subsequent call is silently ignored. Log format and handler behaviour are therefore undefined for most of the application.
2. There is no way to change the log level or output format at runtime without editing source files.

There is also no structured (machine-readable) log output, which makes it difficult to aggregate or query logs in production or CI environments.

### Expected Behaviour

Logger configuration should be centralised in one place. The log level and format should be controllable via environment variables (`LOG_LEVEL`, `LOG_FORMAT`) without any code changes. Setting `LOG_FORMAT=json` should produce one JSON object per line, suitable for log aggregation tools.

### Actual Behaviour

25 modules each call `logging.basicConfig()` with no arguments. Only the first call has any effect. All modules hardcode INFO level. No structured output is available.

```
grep -rn "logging.basicConfig" application/ --include="*.py" | wc -l
# 25
```

### Steps to reproduce

1. Import any two application modules that both call `logging.basicConfig()`.
2. Observe that log format is whatever Python's default is, with no way to override it at runtime.
3. Note that changing `LOG_LEVEL` or `LOG_FORMAT` environment variables has no effect.

shiwani42:
## Summary

Fixes #855

- Add `application/utils/logging_config.py` with a `configure_logging()` function and a `JSONFormatter` class
- `LOG_LEVEL` env var controls log level (default `INFO`)
- `LOG_FORMAT=json` switches to structured JSON output (one object per line), useful for log aggregation
- Remove `logging.basicConfig()` and `logger.setLevel(logging.INFO)` from all 25 affected modules
- Call `configure_logging()` from `create_app()`, `cre_main`, and `worker` so every runtime path is covered

No functional changes to application behaviour. Plain text format remains the default.

## Test plan

- [ ] 10 new tests in `application/tests/logging_config_test.py` covering `JSONFormatter` output, `configure_logging()` level and format selection, env var overrides, and idempotent handler setup
- [ ] All 10 pass: `python -m pytest application/tests/logging_config_test.py`
- [ ] No regressions: `python -m pytest application/tests/db_test.py application/tests/web_main_test.py`
- [ ] `LOG_FORMAT=json` produces valid JSON lines at runtime

### DIFF
--- a/application/__init__.py
+++ b/application/__init__.py
@@ -8,6 +8,7 @@ from flask_sqlalchemy import SQLAlchemy
 from flask_caching import Cache
 from flask_compress import Compress
 from application.config import config
+from application.utils.logging_config import configure_logging
 import os
 import random
 
@@ -26,6 +27,7 @@ cache = Cache()
 
 
 def create_app(mode: str = "production", conf: any = None) -> Any:
+    configure_logging()
     app = Flask(__name__)
     if not conf:
         app.config.from_object(config[mode])
--- a/application/cmd/cre_main.py
+++ b/application/cmd/cre_main.py
@@ -25,10 +25,10 @@ from application.utils import spreadsheet_parsers
 from alive_progress import alive_bar
 from application.prompt_client import prompt_client as prompt_client
 from application.utils import gap_analysis
+from application.utils.logging_config import configure_logging
 
-logging.basicConfig()
+configure_logging()
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
 
 app = None
 
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -47,9 +47,7 @@ from application.utils.gap_analysis import (
 
 from .. import sqla  # type: ignore
 
-logging.basicConfig()
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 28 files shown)


northdpole:
Superseded by https://github.com/OWASP/OpenCRE/pull/1101 — we still want central JSON logging, but as a **standalone** `cre_logging` package so importing a logger never loads Flask/`application.__init__`. Thanks @shiwani42 for the `JSONFormatter` / `configure_logging` direction; the follow-up always emits JSON and includes `module` / `file` / `method`.

northdpole:
Closed as superseded by https://github.com/OWASP/OpenCRE/pull/1101 (`cre_logging` standalone JSON logging).

PR #874, opened 2026-04-08
fix: pin dependencies to exact versions using pip-tools

### LINKED ISSUE #868: Pin dependencies to exact versions in requirements.txt for reproducible builds
### What is the issue?

requirements.txt lists dependencies without pinned versions, meaning pip installs the latest compatible versions each time. This makes the environment non-deterministic: different developers and CI runs can end up with different package versions, leading to inconsistent behavior or unexpected breakages.

(Suggested by @Pa04rth in #854)

### Expected Behaviour

Every `pip install -r requirements.txt` should produce an identical environment regardless of when it is run.

### Actual Behaviour

Dependencies are unpinned (e.g. `Flask`, `click`), so pip resolves to whatever is latest at install time.

### Steps to reproduce

1. Open `requirements.txt` and note entries like `Flask`, `click` have no version pins.
2. Create a fresh virtual environment and run `pip install -r requirements.txt`.
3. Run `pip freeze` and observe that pip has resolved specific versions not specified in `requirements.txt`, showing the environment is determined at install time.

shiwani42:
Closes #868

## Summary

- Introduces pip-tools for reproducible dependency management
- `requirements.in` is the human-maintained list of direct dependencies, deduplicated and normalised from the previous `requirements.txt` (removed 8 duplicates, normalised package names e.g. `scikit_learn` to `scikit-learn`, `pygithub` to `PyGithub`)
- `requirements.txt` is now the pip-compiled lockfile with all transitive dependencies pinned to exact versions, generated with Python 3.12.3 to match CI
- Adds `make update-deps` target for regenerating the lockfile
- Documents the workflow in `docs/developmentSetup.md` so contributors know to edit `requirements.in` and run `make update-deps` rather than editing `requirements.txt` directly

No changes to CI or install workflow. `pip install -r requirements.txt` continues to work as before.

## Test plan

- [x] Fresh venv install from old unpinned `requirements.txt` vs new pinned `requirements.txt`: identical test results (113 tests, same 3 pre-existing failures, same 5 pre-existing errors)
- [x] `black --check` passes on both
- [x] `mypy` produces same 752 pre-existing errors on both
- [x] `flask routes` loads all 41 routes on both
- [x] Lockfile compiled with Python 3.12.3 matching CI


northdpole:
Thanks @shiwani42 for the contribution and for putting time into OpenCRE — please keep contributing when you can.

**Closing:** Superseded by the prod/dev requirements split (#978) for Heroku slug size; we are not adopting pip-tools lockfiles from this branch right now.

PR #879, opened 2026-04-11
Fix weak-link deduplication in gap_analysis: check correct dict level

### LINKED ISSUE #878: Fix: Gap analysis weak-link deduplication checks wrong dictionary level, inflating extra count
### **What is the issue?**

In `application/database/db.py`, the `gap_analysis()` function has a dict-level lookup bug on **line 2224** that breaks weak-link deduplication.

Line 2224 checks `end_key in extra_paths_dict[key]`, but `extra_paths_dict[key]` is `{"paths": {}}` - so `end_key` (a node ID) can never match the only key `"paths"`. The condition is always `False`. Line 2211 (strong-path branch) already does this correctly: `end_key in extra_paths_dict[key]["paths"]`.

### Expected Behaviour

- `grouped_paths[key]["extra"]` should only count unique weak-link destinations
- When multiple paths reach the same destination, the shortest-scoring path should be kept

### Actual Behaviour

- `extra` count is inflated - every weak path increments the counter, even duplicates to the same destination
- The "Show average and weak links (N)" button on the frontend shows a wildly higher N than actual unique destinations
- Deduplication is broken - the last path processed wins instead of the shortest

### Impact

This causes the **"Show average and weak links (N)"** button on the Map Analysis page to display wildly inflated numbers - far higher than the actual number of unique destinations. Every duplicate weak path is counted and stored separately instead of being deduplicated, meaning users see misleading gap analysis results and the shortest-scoring path is not preserved.

**Fix:** `extra_paths_dict[key]` → `extra_paths_dict[key]["paths"]` on line 2224, consistent with line 2211.

PRAteek-singHWY:
## Summary

Fixes #878 

One-line fix for a deduplication bug in `application/database/db.py` → `gap_analysis()`.

## The Bug

Line 2224 checks `end_key in extra_paths_dict[key]` but `extra_paths_dict[key]` is `{"paths": {}}`. The node ID `end_key` can never match the only key `"paths"`, so the condition is always `False`. This breaks weak-path deduplication and inflates the `extra` count shown to users.

The strong-path branch at line 2211 already does this correctly: `end_key in extra_paths_dict[key]["paths"]`.


northdpole:
Reviewed — correct one-line fix. Weak-path dedup now checks `extra_paths_dict[key]["paths"]` matching the strong-path branch. Merging.

PR #880, opened 2026-04-14
fix: remove fragile URL fallback from MyOpenCRE CSV download

PRAteek-singHWY:
### Summary
This PR fixes a bug on the MyOpenCRE page where the CSV download button could fail with a `404 Not Found` error. 

### What Changed
- **Removed** a manual `includes('localhost')` check that would mistakenly overwrite the backend URL to `http://127.0.0.1:5000` (silently stripping away the required `/rest/v1` namespace).
- **Updated** the `fetch` call to simply use the global `${apiUrl}/cre_csv`. 

This ensures the `/rest/v1` prefix is always safely attached, matching how the file upload component right beneath it handles requests.

*Note: Per the `CONTRIBUTING.md` guidelines for quick fixes taking less than 30 minutes, I am submitting this PR directly rather than opening a separate Issue.*


PRAteek-singHWY:
Hi @Pa04rth, this is a quick 1-line routing fix for MyOpenCRE directly as a PR. (So no issue was created as per guidelines)

The CSV download was occasionally losing its /rest/v1 prefix due to an old fallback check, throwing 404s in local dev environments. I removed the check and wired it directly to the global apiUrl so it always routes correctly.

northdpole:
Reviewed — removes fragile localhost fallback that stripped `/rest/v1`. Uses `apiUrl` consistently with upload. Merging.

PR #895, opened 2026-04-29
Fix: Prevent infinite loop in gap_analysis preload (#885)

### LINKED ISSUE #885: Bug: Infinite loop in `gap_analysis.py` preload blocks background runner
### What happened
While looking into `gap_analysis.py`, I found an infinite loop inside the `preload(target_url)` function that can permanently lock up our background processes. 

The function relies on a `while len(waiting):` loop. It expects `calculate_a_to_b(sa, sb)` to eventually return `True`, which allows it to remove a standard pairing from the `waiting` array. However, if a pair continuously fails, it will return `False` forever. The `waiting` array never empties, and the function just sleeps for 30 seconds before endlessly re-trying the same failing standards.

### How it breaks (The Proof)
This isn't just theoretical; the codebase guarantees an infinite loop under two specific scenarios:

1. **Environmental or Heroku 404s:** If `CRE_NO_CALCULATE_GAP_ANALYSIS` is set (or if a standard is missing on Heroku), `/rest/v1/map_analysis` explicitly aborts with a `404`. Over in `preload`, `calculate_a_to_b` sees `status_code != 200`, immediately returns `False`, and goes back to sleep. It will query the exact same 404 endpoint 30 seconds later natively forever.
2. **Failed Background Jobs:** If a heavy `db.gap_analysis` job actually crashes or times out in the RQ queue, its status becomes `FAILED`. When the `preload` loop queries it 30 seconds later, `/rest/v1/map_analysis` sees the failure, creates a *brand new job*, and returns the new `job_id`. Our loop sees the new `job_id`, assumes it's still just "waiting", returns `False`, and falls right back asleep. It gets stuck eternally creating failing jobs.

### Expected Fix
We should implement a `MAX_RETRIES = 10` cap inside the loop. If a specific standard pairing fails repeatedly and crosses the 10-retry threshold, we should drop it from the `waiting` queue so the rest of the application can safely finish.

Also, iterating over the unresolved pairs via Python Tuples instead of repeating the entire nested standards strings array every time will prevent the application from constantly double-checking pairs that have already succeeded.

Let me know if you would like me to open a PR for this, I have a clean fix tested and ready!

PRAteek-singHWY:
### Summary
This PR fixes an infinite loop bug in the `gap_analysis.py` background runner that could permanently lock up the application when checking standard pairings.

### What Changed
* **Added a retry cap:** Implemented a `MAX_RETRIES = 10` threshold so failing pairs are correctly dropped from the waiting list instead of being retried endlessly.
* **Optimized list iteration:** Replaced nested loops over string combinations with a direct list of Python tuples `(sa, sb)` to prevent redundant checking of already succeeded pairs.
* **Refined sleep logic:** Added an `if waiting:` check to safely skip the unnecessary 30-second sleep on the final successful loop iteration.

fixes #885


PRAteek-singHWY:
Hi @Pa04rth, this PR resolves the infinite loop problem reported in #885. The issue has been solved by implementing a 10-retry cap for failing background jobs and optimizing the runner loop with tuples. 
Happy to iterate if needed .

PR #906, opened 2026-05-19
fix(smartlink): redirect to CRE page directly when only one CRE is linked

### LINKED ISSUE #486: Make smartlink go to CRE directly

shreeshtripurwarcomp23-coder:
## Summary
Fixes #486

## What changed
When `smartlink` resolves to a standard node (e.g. CWE/611) that has
exactly **one** connected CRE, the user is now redirected directly to
`/cre/{id}` instead of stopping at the intermediate `/node/...` page.

The CRE page already contains all the information the node page shows,
so the extra click was unnecessary friction.

If zero or multiple CREs are linked, the existing behavior is preserved.

## How it works
In `web_main.py → smartlink()`, after finding the node, we filter
`nodes[0].links` for CRE-typed documents. If exactly one is found,
we extract its `id` and redirect to `/cre/{id}`.

## Tests
- Updated two existing `test_smartlink` assertions to expect the new
  `/cre/{id}` redirect for single-CRE nodes
- Added a new multi-CRE case: a second CRE is linked to the same node,
  confirming the fallback to the node page still works
- All 28 web_main tests pass

### DIFF
--- a/application/tests/web_main_test.py
+++ b/application/tests/web_main_test.py
@@ -571,6 +571,7 @@ class TestMain(unittest.TestCase):
             collection.add_link(dcb, dasvs, ltype=defs.LinkTypes.LinkedTo)
             collection.add_link(dcd, dcwe, ltype=defs.LinkTypes.LinkedTo)
 
+            # single CRE linked via sectionID: should redirect directly to the CRE page (issue #486)
             response = client.get(
                 "/smartlink/standard/CWE/456",
                 headers={"Content-Type": "application/json"},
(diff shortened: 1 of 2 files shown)


shreeshtripurwarcomp23-coder:
@coderabbitai review

Pa04rth (changes requested):
Image
No need of unnecessary concatenation , just stick to the issue and its solution 
Also remove any other unwanted changes you have made other than that is required

PR #908, opened 2026-05-21
fix: always show full list of linked CREs without truncation

### LINKED ISSUE #901: Do not collapse CREs
On a CRE page, lists are now collapsed to prevent long lists - which is good, but that should not be done for CRE's - for example when you go to technical application security controls at https://www.opencre.org/cre/636-660 you only see a small list and then you need to expand it.

Reason for this request: lists of CREs are concise by design and will never be very long, and it's best for navigation if people see all CREs that are contained in another CRE.

shreeshtripurwarcomp23-coder:
fixes #901 

On CRE pages, lists of linked CREs were being collapsed to 5 items
with a "Show more" button. CRE lists are concise by design and
critical for navigation, so they should always be fully visible.

Example page affected: https://www.opencre.org/cre/636-660

Changes:
- Skip truncation when all links in a group are CREs
- Hide "Show more" button for CRE-only groups
- Non-CRE links (Standards, Tools) still truncate as before
- 13 unit tests added

### DIFF
--- a/application/frontend/src/components/DocumentNode/DocumentNode.test.ts
+++ b/application/frontend/src/components/DocumentNode/DocumentNode.test.ts
@@ -1,0 +1,136 @@
+/**
+ * Tests for the fix: CRE-type linked documents should never be truncated/collapsed.
+ *
+ * Issue: On a CRE page, lists of linked CREs were being collapsed after MAX_LENGTH_FOR_AUTO_EXPAND (5)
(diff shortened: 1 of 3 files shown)


Pa04rth:
@shreeshtripurwarcomp23-coder We dont need a test for this small thing , and also could you please explain the need of jest.unit.config file ??

PR #909, opened 2026-05-24
fix: always show all CRE links on CRE pages without collapsing (#901)

### LINKED ISSUE #901: Do not collapse CREs

shreeshtripurwarcomp23-coder:
Fixes #901

## Problem
On a CRE page, lists of linked CREs were being collapsed after 5 items,
requiring users to click "Show more" to see the rest. This happened because
the collapse logic in CommonRequirementEnumeration.tsx did not distinguish
between CRE links and links to external standards.

## Fix
Added a check in CommonRequirementEnumeration.tsx: if all links in a group
are CREs, show all of them without collapsing and hide the "Show more" button.
Links to external standards still collapse as before.

This matches the logic already present in DocumentNode.tsx (allLinksAreCres).

## Changes
- `application/frontend/src/pages/CommonRequirementEnumeration/CommonRequirementEnumeration.tsx`
  - Import `DOCUMENT_TYPES` from `../../const`
  - Compute `allLinksAreCres` before rendering each link group
  - Use `visibleResults` (full list when CREs, sliced otherwise) instead of inline slice
  - Guard the "Show more" button with `!allLinksAreCres`

### DIFF
--- a/application/frontend/src/pages/CommonRequirementEnumeration/CommonRequirementEnumeration.tsx
+++ b/application/frontend/src/pages/CommonRequirementEnumeration/CommonRequirementEnumeration.tsx
@@ -8,6 +8,7 @@ import { Icon } from 'semantic-ui-react';
 import { DocumentNode } from '../../components/DocumentNode';
 import { ClearFilterButton, FilterButton } from '../../components/FilterButton/FilterButton';
 import { LoadingAndErrorIndicator } from '../../components/LoadingAndErrorIndicator';
+import { DOCUMENT_TYPES } from '../../const';
 import { useEnvironment } from '../../hooks';
 import { applyFilters, filterContext } from '../../hooks/applyFilters';
 import { Document } from '../../types';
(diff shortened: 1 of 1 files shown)


Pa04rth:
@shreeshtripurwarcomp23-coder  Add the screenshot/screen recording of the result

shreeshtripurwarcomp23-coder:
@Pa04rth 
Thank you for the feedback, @Pa04rth !
Please find the screenshot attached showing the fix working locally.

The issue was in `CommonRequirementEnumeration.tsx` — the collapse logic (slicing the list after 5 items and showing a "Show more" button) was being applied to all link types including CREs, with no distinction between CRE links and links to external standards.

The fix adds an `allLinksAreCres` check that mirrors the same pattern already correctly implemented in `DocumentNode.tsx`. When all links in a group are of type CRE, the full list is shown without slicing and the "Show more" button is hidden. Links to external standards continue to collapse as before.

Changes are limited to a single file: `CommonRequirementEnumeration.tsx`
- Added `DOCUMENT_TYPES` import from `../../const`
- Compute `allLinksAreCres` before rendering each link group
- Use `visibleResults` (full list for CREs, sliced for others) instead of inline slice
- Guard the "Show more" button with `!allLinksAreCres`

OWASP SCREENSHOT

Pa04rth (approved):

Pa04rth:
@shreeshtripurwarcomp23-coder Approved

shreeshtripurwarcomp23-coder:
@Pa04rth 
I've force pushed a signed/verified commit, could you please approve the workflows to run?

PR #920, opened 2026-06-07
GSOC-Week1-Module_A

ParthAggarwal16:
# Summary

This PR adds the initial repository configuration loading and validation layer for the harvester pipeline.

## Added

- YAML-based repository configuration loading
- Pydantic schema validation for repository configs
- Semantic repository validation layer
- Validation fixtures and pytest coverage
- Sample harvester repository configuration
- Repository validation utilities and exports
- Exclude pattern support scaffold

### Config Loader Validation
- Missing repository ID
- Invalid chunk size
- Invalid YAML syntax
- Missing config file
- Empty owner validation
- Invalid chunking strategy
- Invalid polling mode

### Repository Semantic Validation
- Duplicate repository IDs
- Duplicate repositories
- Duplicate include paths

## Test Plan

Executed:

bash python3 -m pytest application/tests/harvester_test -v python3 -m coverage report python3 -m black application/utils/harvester application/tests/harvester_test 

All tests passing.

## Notes for Reviewers

- Validation is intentionally split into:
  - schema validation (config_loader.py)
  - semantic/business-rule validation (repos_validator.py)

- Repository uniqueness checks are case-insensitive.

- exclude_patterns.txt is currently scaffolded for future integration into harvesting logic.

image

https://github.com/ParthAggarwal16/OpenCRE/pull/3#issuecomment-4586752992
this comment has some good coderabbit diagrams if you guys wanna have a look at it


ParthAggarwal16:
https://github.com/ParthAggarwal16/OpenCRE/pull/3#issuecomment-4586736028
also this one has the entire HLA for more context

ParthAggarwal16:
@coderabbitai
[0be6df8](https://github.com/OWASP/OpenCRE/pull/920/commits/0be6df87c7642908c76dc915c48b54c30227acff)
this commit should address it

northdpole (changes requested):
Great job on the W1 config layer — the schema vs semantic validation split (`config_loader.py` / `repos_validator.py`) is clean, the Pydantic models are well-structured, and the fixture matrix covers the important failure modes. This is ready to merge once the items below are addressed to match repo conventions.

## Requested changes

### 1. Make tests discoverable by `make test` (blocker)

CI runs:

```
python -m unittest discover -s application/tests -p "*_test.py"
```

Your tests are named `test_config_loader.py` / `test_repos_validator.py` and use pytest — they never run in CI today.

**Fix:**
- Rename to `config_loader_test.py` and `repos_validator_test.py` (under `application/tests/harvester_test/`).
- Convert to `unittest` (mirror `application/tests/noise_filter/config_loader_test.py`): `unittest.TestCase` classes, `self.assertRaises` / `self.assertEqual`, no `pytest.raises`.
- Run locally: `make test` and confirm both modules are executed (not just `flask routes`).

### 2. Cover the remaining fixtures

You already have fixtures without tests:
- `invalid_polling_interval.yaml` (`interval_minutes: 0`) — add a test expecting `ConfigLoaderError` (Pydantic `gt=0` on `interval_minutes`).
- `empty_include_paths.yaml` (`include: []`) — add a test expecting `ConfigLoaderError` (`min_length=1` on `paths.include`).

### 3. Rebase onto `main`

The branch is ~33 commits behind `main`. Please rebase and push so CI runs fresh on current `main`.

## Note for upcoming weeks (not blocking W1)

Module B’s `ChangeRecord` contract (merged in #913) expects:
- `source.repo` as a combined slug, e.g. `"OWASP/ASVS"` (not separate `owner` + `repo` fields in the emitted record).
- Chunk sizing in **characters** (`max_chars`, default 4000 in the labeled-dataset builder), while this config uses `max_tokens`.

Fine for W1 config-only work — please align chunking units and emission shape in A-W3 so harvest output matches the B contract without a surprise refactor.

## Production

No concerns — this is additive and nothing is wired into web/worker/import yet.

Ping when the test/convention fixes are in and rebased; happy to approve and merge then.

ParthAggarwal16:
yo @northdpole , sorry it took a bit longer than enxpected, but i think it should be good to go now, let me know if anything else is neccesery or i missed
image

application/utils/harvester/exclude_patterns.txt
Pa04rth, line 1: add .claude and .cursor files

ParthAggarwal16:
yeah mb, forgot about that one, should be done now

application/utils/harvester/config_loader.py
```diff
@@ -0,0 +1,28 @@
...
+class ConfigLoaderError(Exception):
+    # when repo config loading fails
+    pass
+
+
+def load_repo_config(path: str | Path) -> ReposFile:
+    config_path = Path(path)
+
+    if not config_path.is_file():
+        raise FileNotFoundError(f"Configuration file not found: {config_path}")
```
Pa04rth: I think this error should also be  ConfigLoadError() , because rest two errors are handled gracefully but this one cause direct termination . In future if we trigger a cron job to retry after getting this error , we may apply to all three of them 

or we can do this -- 
```
class ConfigLoaderError(Exception):
    """Base class for all configuration loading errors."""
    pass


class ConfigFileNotFoundError(ConfigLoaderError):
    """Raised when the configuration file cannot be found."""
    pass
```

application/utils/harvester/exclude_patterns.txt
Pa04rth, line 1: It seems like this is currently a dead file , please specify in the comments if this is loaded in the next PRs

application/utils/harvester/schemas.py
```diff
@@ -0,0 +1,104 @@
...
+    strategy: Literal["markdown", "plaintext"] = Field(
+        ...,
+        description="Chunking strategy used for text segmentation",
+    )
+
+    max_tokens: int = Field(..., gt=0, description="max token size per chunk")
+
+    overlap_tokens: int = Field(
+        ge=0,
+        default=100,
```
Pa04rth: Thinking of one condition here , which can fail , what if user put max_tokens = 100 and does not specify overlap_token , then the default value will be used which will result in a misleading error . 

There can be 3 fixes -- 
1- Overlap_token value should depend on max_token value
2- Make the Overlap _token Value to be small like 20 or 30 
3- Make Overlap_token compulsory to mention

application/utils/harvester/schemas.py
```diff
@@ -0,0 +1,104 @@
...
+    exclude: list[str] = Field(
+        default_factory=list, description="Glob patterns to exclude during ingestions"
+    )
+
+
+# this will define how the harvested data should be chunked before downstream
+class ChunkingConfig(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    strategy: Literal["markdown", "plaintext"] = Field(
```
Pa04rth: In RFC the defines **chunking.strategy** as `markdown_heading | html_readability | fixed_size.` This PR's **ChunkingConfig.strategy**  is `Literal["markdown", "plaintext"]` . Please make appropriate changes/comments mentioning the change in the RFC so the next module programmer didn't face any issue .

Pa04rth (changes requested):

application/utils/harvester/schemas.py
ParthAggarwal16: went for option 2, mainly because :
in option 1, pydantic defaults depending on another field can become annoying
in option 3, i think default should exist for convenience,

application/utils/harvester/exclude_patterns.txt
ParthAggarwal16: specified in the file now, it will be in the week4 PR

ParthAggarwal16:
[f11f869](https://github.com/OWASP/OpenCRE/pull/920/commits/f11f869f716dbca9b0d5ec5fd86f2729cfc4aed4)

@Pa04rth this one should address all the concerns

Pa04rth (approved):

Pa04rth:
@ParthAggarwal16  Good work , let @northdpole approve the PR , then we can merge

application/utils/harvester/repos.yaml
```diff
@@ -0,0 +1,43 @@
...
+    branch: master
+    paths:
+      include:
+        - "4.0/en/**/*.md"
+
+      exclude:
+        - "**/archive/**"
+
+    chunking:
+      strategy: markdown
```
northdpole: **Blocker:** `markdown` is not accepted by `ChunkingConfig`, so loading the packaged `repos.yaml` raises `ConfigLoaderError`. Apply the fix in this PR—not only in stacked #983—by changing this value and the second repository value on line 37 to `markdown_heading`. Then add a test that calls `load_repo_config()` on `application/utils/harvester/repos.yaml` and asserts both repository entries load successfully. This ensures every intermediate merge leaves `main` usable.

application/utils/harvester/schemas.py
```diff
@@ -0,0 +1,104 @@
...
+    mode: Literal["full", "incremental"] = Field(
+        ..., description="repository sync mode"
+    )
+
+    interval_minutes: int = Field(..., gt=0, description="polling interval in minutes")
+
+
+# top level repository ingestion configuration
+class RepositoryConfig(BaseModel):
+    model_config = ConfigDict(extra="forbid")
```
northdpole: `min_length=1` still accepts whitespace-only values for `id`, `owner`, `repo`, and `branch`, which then fail later in cache/Git operations. Apply the fix at the model level by changing this configuration to strip string whitespace before field validation (for example, `ConfigDict(extra="forbid", str_strip_whitespace=True)`) while retaining each `min_length=1`. Add parameterized/focused tests proving `" "` is rejected for all four fields. If GitHub-only sources are the intended contract, also constrain owner/repository names to valid GitHub syntax.

application/tests/harvester_test/config_loader_test.py
```diff
@@ -0,0 +1,63 @@
...
+        config_path = FIXTURES_DIR / "invalid_polling_interval.yaml"
+
+        with self.assertRaises(ConfigLoaderError):
+            load_repo_config(config_path)
+
+    def test_empty_include_paths(self):
+        config_path = FIXTURES_DIR / "empty_include_paths.yaml"
+
+        with self.assertRaises(ConfigLoaderError):
+            load_repo_config(config_path)
```
northdpole, line 59: Please extend this test class to exercise the validation paths represented by the existing but unused `empty_owner.yaml`, `invalid_chunking_strategy.yaml`, and `invalid_polling_mode.yaml` fixtures; each should call `load_repo_config()` and assert `ConfigLoaderError`. Also add cases for whitespace-only repository fields and `overlap_tokens >= max_tokens`, plus a success test loading the real packaged `repos.yaml`. In `repos_validator_test.py`, add case-only duplicate IDs/repositories so the intended `casefold()` behavior is locked down.

ParthAggarwal16:
the changes are addressed, good to go from my side

northdpole (approved):
Re-reviewed after your Jul 19 fixes. The earlier blockers are addressed:

- packaged `repos.yaml` uses `markdown_heading`
- `str_strip_whitespace=True` + whitespace-only field tests
- unused fixtures now exercised in `config_loader_test.py`
- tests are `*_test.py` / unittest and CI is green

Approving and merging.

PR #928, opened 2026-06-12
GSoC 2026 Module B — Week 2: Stage 1 regex filter + Stage 1.5 sanitize

manshusainishab:
## Summary

- Adds **Stage 1** (path-based regex/glob noise filter) and **Stage 1.5** (defensive text sanitization vendored from TRACT) of the Module B pipeline.
- Addresses two of the six CodeRabbit comments on the Week 1 PR (#4 and #5 — chunker `<pre>` tracking and hard-split cursor arithmetic).

Builds on Week 1 (`module_b_w1`, in review). Once Week 1 lands, this PR will rebase against `main` automatically.

Part of the GSoC 2026 OpenCRE Scraper & Indexer (Project OIE) Module B (Noise/Relevance Filter).

### Stage 1.5 — text sanitization (commit `77ad228`)

| File | Purpose |
|---|---|
| `application/utils/noise_filter/sanitize.py` | `sanitize_text(text)` + `strip_html(text)`. Defensive cleanup: PDF ligatures (`ﬀ → ff`), zero-width chars, HTML entities/tags, broken hyphenation, NFC normalization. Idempotent on clean Module A output. |
| `application/tests/noise_filter/sanitize_test.py` | 26 unittest cases — pipeline rules in isolation, idempotency across input classes, edge cases, public helper. |

**Vendored from [rocklambros/TRACT](https://github.com/rocklambros/TRACT/blob/main/tract/sanitize.py)** (CC0 1.0 Universal), with one deliberate deviation: we do NOT collapse interior whitespace. TRACT's pipeline ends with `re.sub(r"\s+", " ", text)` which flattens newlines too — that's right for embedding-similarity use cases but destroys structure Module A's contract explicitly preserves (rule 6: whitespace inside code fences). Module B's LLM benefits from paragraph breaks and code-fence layout, so we keep them intact. Also dropped TRACT's `max_length`/`return_full` machinery and `sanitize_control(dict)` helper — both belong in their own domain, not B's.

### Stage 1 — path-based regex filter (commit `8f5169b`)

| File | Purpose |
|---|---|
| `application/utils/noise_filter/regex_filter.py` | `RegexFilter` class. Public surface: `is_noise_path(path)`, `is_noise_record(rec)`, `filter_records(records)` lazy generator. |
| `application/utils/noise_filter/noise_patterns.yaml` | Data-driven blocklist: extensions, filenames, path globs, allow-overrides. Contributors edit this file (not code) to extend coverage. |
| `application/tests/noise_filter/regex_filter_test.py` | 15 unittest cases including the plan's acceptance criteria: **≥90% rejection** on a table of known junk paths, **0% false positives** on a table of known security doc paths. |

**Conservative-by-design** under the recall-first labeling rule (agreed with the maintainer on 2026-06-01, see the Week 1 PR thread). Stage 1 only blocks paths where we're highly confident no security content lives there. We **deliberately do NOT block** `**/blog/**`, `Website/content/**`, or `**/sponsors.md` even though Week 1's labeling found them ~100% organizational — recall-first says let Stage 2 LLM judge content rather than blocking at the path level. We DO block `**/Supporting Resources/meetings/**` and `**/Supporting Resources/enterprise metrics/**` because Week 1 found those 100% organizational across all 25 SAMM samples.

### CodeRabbit Week 1 deferrals (commit `806b0e5`)

Addresses two CodeRabbit comments on the Week 1 PR that were deferred to Week 2 to land alongside other harvester work:

- **#4**: chunker now tracks `<pre>` block depth in addition to ```` ``` ```` fences, so a heading-like line inside `<pre>...</pre>` no longer triggers a false split.
- **#5**: `_split_chunk_by_size` cursor arithmetic now records per-entry separator width (0 for hard-split fragments, 2 for normal `\n\n` boundaries) instead of unconditionally adding `+2` between every consecutive sub-chunk.

**Empirical check** on the existing Week 1 dataset: 0/100 chunks contain `<pre>` tags and 0/100 are near the 4000-char hard-split ceiling. So the existing `labeled_data.json` and `candidate_commits.json` are unchanged — both old and new chunker produce bit-identical output on our sample. The fixes are forward-looking, protecting future harvests from corner-case files we didn't happen to sample.

## Test plan

- [x] `make test` — **312 passing, 1 skip, 0 failures, 0 errors** (was 271 after Week 1; +41 from Week 2: 15 regex_filter + 26 sanitize).
- [x] `black --check .` — clean across the whole repo (164 files unchanged, 0 to reformat). Pinned 24.4.2, matching superlinter.
- [x] Plan acceptance criteria for Stage 1: ≥90% rejection on table of known junk, 0% false positives on table of known security docs.
- [x] Smoke tests on the chunker fixes confirm `<pre>` tracking prevents false splits and hard-split sub-chunks are contiguous (no phantom +2).

## Notes for reviewers

- **Out of this PR (intentional):**
  - Stage 2 LLM classifier wrapping `PromptHandler` (LiteLLM-backed) — Week 3 deliverable.
  - End-to-end pipeline wiring (Stage 1 → Stage 1.5 → Stage 2) — Week 5 deliverable.
  - `KnowledgeQueueItem` SQLAlchemy model + Alembic migration — Week 5 deliverable.
- **Branch base:** branched off `module_b_w1` so that `ChangeRecord` from Week 1 is available. PR targets `main` directly; GitHub will rebase once Week 1 lands.
- **Labeling rule (recall-first):** under maintainer agreement on 2026-06-01, KNOWLEDGE = any chunk with a security signal; NOISE = only organizational/admin content. Week 2's Stage 1 patterns are aligned with this rule — they err on the side of letting through.

Note :- Claude is used for some parts of the PR and file changes.


application/utils/noise_filter/schemas.py
```diff
@@ -0,0 +1,160 @@
...
+    a commit); future schemes may include "feed_post" for RSS or others.
+    `id` is the unique identity within the scheme; `path` is a convenience
+    duplicate for repo_path (id == path in practice).
+    """
+
+    model_config = ConfigDict(extra="ignore", str_strip_whitespace=True)
+
+    kind: str = Field(min_length=1)
+    id: str = Field(min_length=1)
+    path: str = Field(min_length=1)
```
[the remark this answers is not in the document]
manshusainishab: Acknowledged. Deferring the Locator discriminated-union refactor to a follow-up PR — it's a "heavy lift" that requires regenerating the JSON Schema artifact, updating the contract documents, and adjusting tests. Today the schema correctly validates what Module A actually emits (only kind=repo_path is observed). When RSS records start flowing through in a future Module A version, we'll land the RepoPathLocator/FeedPostLocator split in the same PR that introduces the first RSS test fixtures.

application/utils/noise_filter/noise_patterns.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - .min.js
+  - .min.css
+  - .woff
+  - .woff2
+  - .ttf
+  - .eot
+  - .pdf   # PDFs reach Stage 2 via Module A as extracted text; we never see the binary
+
+# --- Tier 1: filenames that are universally non-security ------------------
+deny_filenames:
```
Pa04rth, line 41: I think we must add AI coding agents filenames here too like .claude or .cursor  and we may encounter them and they are also noise  ??

application/utils/noise_filter/noise_patterns.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+# --- Allow-overrides ------------------------------------------------------
+# If a path matches both a deny rule and an allow_override, the override wins.
+# Lets us extend deny coverage without breaking valid documents.
+#
+# Pattern note: `docs/**` covers both top-level (e.g. docs/CONTRIBUTING.md)
+# AND nested (e.g. docs/designs/foo.md) paths. `docs/**/*.md` would NOT
+# match top-level files because fnmatch requires the trailing `/*` to consume
+# at least one chunk of name -- so we use the broader form.
+allow_overrides:
```
Pa04rth, line 104: The comment -- "# OpenCRE's own docs are always allowed " is repository-specific, but the rule is repository-agnostic.

I think this is better 
allow_overrides:
  - "OpenCRE/docs/**"

Happy to discuss

Pa04rth (commented):
Please address the comments , rest is great

application/utils/noise_filter/noise_patterns.yaml
manshusainishab: Thanks @Pa04rth — both good catches. Addressed in 993d016.

Moved .claude, .cursor to deny_paths rather than deny_filenames. They are typically directories (.claude/settings.json, .cursor/rules/), so they belong alongside .github/**. Same noise category, just the right structural bucket. Also added .aider/** and .continue/** while in there.

application/utils/noise_filter/noise_patterns.yaml
manshusainishab: Updated the allow_overrides comment instead of changing the pattern to OpenCRE/docs/**. Per Module A's input contract v0.3, locator.path is repo-relative POSIX (no OWASP/<repo>/ prefix), so a literal OpenCRE/docs/** pattern would match nothing in practice — the repo identity lives in source.repo, not in the path. The original rule docs/** is intentionally repo-agnostic: multiple OWASP projects keep narrative content in docs/ and we want all of them to reach Stage 2 LLM judgment. New comment makes that intent explicit.

If true repo-scoped overrides become a need (e.g. one repo's docs/ is all noise but another's is gold).

Pa04rth (approved):

PR #939, opened 2026-06-18
cwe: add XXE search aliases and enforce prohibited CWE exclusions

Bornunique911:
This PR builds on the earlier CWE mapping work (PR #823 ) and addresses two gaps found during local verification after the previous merge.

Problem:
Searching for terms like 'XXE' did not reliably surface the expected top-level CWE result, even when the corresponding CWE data existed locally. This created a mismatch between the intended curated CWE import behavior and actual API/UI discoverability.

Solution:
- Expand text-search aliases to include common security shorthand terms (e.g., XXE), ensuring related CWE and CRE records surface correctly.
- Strengthen CWE import filtering to consistently exclude prohibited entries.

Scope:
- CWE import filtering logic
- Search alias expansion for XXE and related terms
- Unit tests covering prohibited CWE exclusion and XXE search behavior
- Script messaging update for the CWE refresh workflow

Testing:
```
./venv/bin/python -m pytest application/tests/cwe_parser_test.py -k 'register_CWE or prohibited' -q 
```

```
./venv/bin/python -m pytest application/tests/db_test.py -k 'text_search' -q
```

Follows-up: #823 
References: #472

PR #941, opened 2026-06-20
Fix: Narrow except Exception to except SQLAlchemyError in health_check()

### LINKED ISSUE #940: Fix overly broad exception handling in Node_collection.health_check to catch SQLAlchemyError only
The `health_check()` method in `db.py` currently uses a generic `except Exception` as a fallback after catching `OperationalError`. This can mask programming faults (e.g., AttributeError, TypeError) and make debugging harder. The fallback handler should catch only database‑layer exceptions that are expected when the database is unhealthy, specifically `SQLAlchemyError`.

## Expected Behavior
Database connectivity issues should be reported as `ok=False` with appropriate reason.

Other errors (e.g., NameError, misused attribute) should not be caught silently; they should propagate to reveal bugs.

The fallback handler should only suppress SQLAlchemy‑specific exceptions` that are not `OperationalError` (though SQLAlchemyError covers them, we already catch OperationalError separately).

## Actual Behavior
Any unexpected exception is caught and returns `{"ok": False, "db_reachable": False, ...}`, making it difficult to detect programming errors in the health check logic.

## Proposed Fix
Add `SQLAlchemyError` to the import list from `sqlalchemy.exc`.

Replace `except Exception:` with `except SQLAlchemyError:`.

## Updated code:

```python
from sqlalchemy.exc import OperationalError, IntegrityError, SQLAlchemyError

# ... inside health_check ...
except SQLAlchemyError:  # pragma: no cover
    return {
        "ok": False,
        "db_reachable": False,
        "reason": "database health query failed",
    }
```
### Reference
- This fix has been implemented in PR: #939

Bornunique911:
## Summary
Fixes overly broad exception handling in `Node_collection.health_check()` by replacing `except Exception` with `except SQLAlchemyError`.

Closes #940

## Problem
The `health_check()` method in `application/database/db.py` currently uses a generic `except Exception` as a fallback after catching `OperationalError`:

```python
except OperationalError:
    return {"ok": False, "db_reachable": False, "reason": "database unreachable"}
except Exception:  # pragma: no cover - defensive, never fail open
    return {"ok": False, "db_reachable": False, "reason": "database health query failed"}
```

This broad catch silently swallows programming bugs such as AttributeError, TypeError, or NameError within the health check logic itself. Instead of propagating the error (which would reveal the bug), it returns a structured {"ok": False} response, making such issues very hard to detect.

## Solution
- Add `SQLAlchemyError` to the import from `sqlalchemy.exc`

- Replace `except Exception:` with `except SQLAlchemyError:`.

Since OperationalError is a subclass of SQLAlchemyError and is caught first, the new clause only handles remaining SQLAlchemy-level errors.

## Changes
`application/database/db.py`

Change 1 – Import (line ~26):

```diff
- from sqlalchemy.exc import OperationalError, IntegrityError
+ from sqlalchemy.exc import OperationalError, IntegrityError, SQLAlchemyError
```
Change 2 – Exception handler (lines ~2292–2296):

```diff
- except Exception:  # pragma: no cover - defensive, never fail open
+ except SQLAlchemyError:
      return {
          "ok": False,
          "db_reachable": False,
          "reason": "database health query failed",
      }
```

## Behavioral Change
⚠️ Important: Non-SQLAlchemy exceptions (e.g., programming bugs) will now propagate naturally rather than being swallowed. This is intentional — bugs should surface during development/testing, not be hidden behind a structured health-check response.

In production, this means:

Before: Any unexpected error → `{"ok": False, ...}` (bug hidden)

After: SQLAlchemy errors → `{"ok": False, ...}`; programming bugs → HTTP 500 with stack trace (bug visible)

## Related Issue
Closes #940

### DIFF
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -2290,11 +2290,11 @@ class Node_collection:
                 "reason": "database unreachable",
             }
         except SQLAlchemyError:  # pragma: no cover - defensive, never fail open
-             return {
-                 "ok": False,
-                 "db_reachable": False,
-                 "reason": "database health query failed",
-             }
+            return {
+                "ok": False,
+                "db_reachable": False,
+                "reason": "database health query failed",
+            }
 
         if cre_count == 0 or standards_count == 0:
             return {
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -23,7 +23,7 @@ from sqlalchemy.orm import aliased
 from flask_sqlalchemy.model import DefaultMeta
 from sqlalchemy import func, delete, cast as sql_cast, literal, or_
 from sqlalchemy.dialects.postgresql import JSONB
-from sqlalchemy.exc import OperationalError, IntegrityError
+from sqlalchemy.exc import OperationalError, IntegrityError, SQLAlchemyError
 
 from neomodel import (
     config,
@@ -2289,12 +2289,12 @@ class Node_collection:
                 "db_reachable": False,
                 "reason": "database unreachable",
             }
-        except Exception:  # pragma: no cover - defensive, never fail open
-            return {
-                "ok": False,
-                "db_reachable": False,
-                "reason": "database health query failed",
-            }
+        except SQLAlchemyError:  # pragma: no cover - defensive, never fail open
+             return {
+                 "ok": False,
+                 "db_reachable": False,
+                 "reason": "database health query failed",
+             }
 
         if cre_count == 0 or standards_count == 0:
             return {


Bornunique911:
@coderabbitai review

northdpole:
Thanks — the `health_check` narrowing is correct.

This hunk is already included in [#939](https://github.com/OWASP/OpenCRE/pull/939) (commit `dbea8ba9` on that branch), which is approved for merge. Recommend closing this PR as superseded so we do not land the same 2-line change twice.

No action needed on your side; the fix will ship via #939.

northdpole:
Superseded by #939 (merged).

PR #943, opened 2026-06-21
feat(frontend): add reusable useUser hook and header auth UX

### LINKED ISSUE #942: Login MVP: add reusable user/auth state + header login/logout UX
Part of the Login MVP work (priority #1 on @northdpole ). Scoped to the non-MyOpenCRE auth UX piece per discussion with @Pa04rth  — the MyOpenCRE/capabilities side is owned separately by Prateek and is intentionally out of scope here.

Current state: Google OAuth is fully wired backend-side (/rest/v1/login, /callback, /logout, /rest/v1/user), but the only place that consumes it is chatbot.tsx, with the auth-check logic inlined and written to force login (it redirects unauthenticated users to the OAuth flow). There's no shared way to read login state, and the header doesn't reflect whether a user is logged in.

Proposal:

Add a reusable useUser hook that probes /rest/v1/user and reports login state. /rest/v1/user is @login_required and returns 401 for anonymous users, so the hook treats 401 as "not logged in" — without redirecting, so public pages stay accessible to anonymous visitors (consistent with the public-reads-stay-open direction).
Surface auth state in the header: show the logged-in user + a logout action, or a login button when anonymous.

Out of scope (deliberately): any backend change, the myopencre/capabilities flag, cre_csv_import gating, and refactoring chatbot to use the hook (can follow later). Frontend-only, kept small.
PR to follow.

skypank-coder:
Scoped to non-MyOpenCRE auth UX per @Pa04rth 's steer (b): a reusable useUser hook + header login/user/logout state, following the existing chatbot auth pattern. Part of the Login MVP (@northdpole 's priority #1).

useUser probes /rest/v1/user and treats 401 as anonymous — no forced redirect, so public pages stay open to anonymous users (consistent with the public-reads-stay-open / v2 direction @<spyros-handle> outlined).
Header shows the logged-in user + Logout, or a Login button when anonymous.
Frontend-only. No backend changes. No myopencre/capabilities coupling (left to Prateek).

Verified manually in dev: NO_LOGIN=1 shows logged-in state; unauthenticated shows Login with no redirect on load; login/logout flows work.
Closes #942 .

### DIFF
--- a/application/frontend/src/hooks/index.ts
+++ b/application/frontend/src/hooks/index.ts
@@ -1,3 +1,4 @@
 export { useEnvironment } from './useEnvironment';
 export { useLocationFromOutsideRoute } from './useLocationFromOutsideRoute';
 export { useCapabilities } from './useCapabilities';
+export { useUser } from './useUser';
--- a/application/frontend/src/hooks/useUser.ts
+++ b/application/frontend/src/hooks/useUser.ts
@@ -20,17 +20,25 @@ export const useUser = () => {
         if (res.status === 200) {
           return res.text();
         }
-        return null; // 401 or anything else => treated as not logged in
+        if (res.status === 401) {
+          return null; // the normal anonymous case — not logged in
+        }
+        // Unexpected status (e.g. 5xx): a real failure, not a clean anonymous state.
+        throw new Error(`Unexpected /user status: ${res.status}`);
       })
       .then((value) => {
         if (active) {
           setUser(value && value.trim() !== '' ? value : null);
         }
       })
-      .catch(() => {
+      .catch((err) => {
+        // Network error or unexpected status. Degrade to anonymous so public
+        // pages stay accessible (never block or redirect), but log the failure
+        // instead of silently masking it as a normal logged-out state.
         if (active) {
-          setUser(null); // network error => treat as anonymous, do NOT redirect
+          setUser(null);
         }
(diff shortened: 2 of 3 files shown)


northdpole (changes requested):
Direction looks good — `useUser` treating 401 as anonymous (no redirect) is what we want.

One ask before merge: **gate the header Login/Logout behind `capabilities.login`**, same pattern as `capabilities.myopencre`. We want to roll out auth UI per-environment; prod should be able to hide it with `login: false` until we're ready.

`useUser` can stay as-is — just wrap the header auth blocks in `capabilities.login && …`. Default `login: false` when `/api/capabilities` is missing/fails is fine for now.

Capabilities endpoint wiring can be a small follow-up if you prefer, but please don't merge always-on header auth without the gate.

skypank-coder:
@northdpole Pushed updates addressing both review comments .

northdpole (approved):
Rebased onto main after #949. Kept main's `useCapabilities` normalization; Header auth UI gated on `capabilities.login` as requested. `useUser` treats 401 as anonymous with no redirect — good for public-read pages.

northdpole:
Thanks @skypank-coder — clean work on the Login MVP. The `useUser` hook (401 → anonymous, no redirect) and gating Header auth behind `capabilities.login` are exactly what we wanted for a safe per-environment rollout. Appreciate the quick follow-ups on review feedback too.

PR #947, opened 2026-06-25
GSoC 2026 Module B — Week 3: Stage 2 LLM relevance classifier

manshusainishab:
## Summary

Adds Stage 2 of Module B (Noise/Relevance Filter): an LLM classifier that labels
each content chunk as **KNOWLEDGE**, **NOISE**, or **UNCERTAIN** under the
recall-first rule. Builds on the Week 1 schemas and Week 2 regex/sanitize stages.

This PR is self-contained (classifier + prompt + config + tests). Pipeline
wiring, the queue/DB model, and the CLI entry point come in later weeks.

## What's added

- **`application/utils/noise_filter/config_loader.py`** — loads Module B settings
  from `CRE_NOISE_FILTER_*` environment variables into a typed `NoiseFilterConfig`
  (model, batch size, per-chunk char cap, confidence threshold), with defaults.
- **`application/utils/noise_filter/prompts.py`** — the recall-first system prompt
  and a few-shot block (5 KNOWLEDGE / 3 NOISE / 2 UNCERTAIN worked examples), plus
  a helper that renders a numbered batch of chunks into the user prompt.
- **`application/utils/noise_filter/llm_classifier.py`** — `LLMClassifier`, which
  classifies a list of `ChangeRecord`s and returns one `ClassifyResult` per record.
- **`application/tests/noise_filter/llm_classifier_test.py`** — 14 unit tests,
  fully mocked (no network calls).
- **`.env.example`** — documents the four new `CRE_NOISE_FILTER_*` variables.

## How the classifier works

- Sends each chunk's `heading_path` + `text` to a dedicated lightweight model via
  LiteLLM (default `gemini/gemini-2.5-flash-lite`).
- Processes records in batches (`CRE_NOISE_FILTER_BATCH_SIZE`, default 10), one
  request per batch, and maps results back to input order by index.
- Requests a strict JSON-schema response; if the provider doesn't support strict
  mode, falls back to JSON-object mode.
- Truncates each chunk to `CRE_NOISE_FILTER_MAX_CHARS` (default 1500) before sending.
- Retries on rate-limit/quota errors using the existing `CRE_LLM_MAX_RETRIES` /
  `CRE_LLM_RETRY_SLEEP_SECONDS` settings.
- Returns `UNCERTAIN` (confidence 0.0) for any unparseable, malformed, or invalid
  output, and marks a whole batch `UNCERTAIN` if the LLM call fails — so a bad
  response never aborts a run.

## Configuration

| Variable | Default | Purpose |
|---|---|---|
| `CRE_NOISE_FILTER_LLM_MODEL` | `gemini/gemini-2.5-flash-lite` | Classification model (LiteLLM string) |
| `CRE_NOISE_FILTER_BATCH_SIZE` | `10` | Chunks per LLM request |
| `CRE_NOISE_FILTER_MAX_CHARS` | `1500` | Per-chunk character cap before sending |
| `CRE_NOISE_FILTER_CONFIDENCE_THRESHOLD` | `0.8` | Minimum confidence to enqueue a KNOWLEDGE verdict |

The model is Gemini, so it authenticates with the existing `GEMINI_API_KEY`;
no new credential is required.

## Testing

- 14 new unit tests covering prompt content, batch ordering and splitting,
  malformed/invalid/empty output handling, the JSON-schema fallback, rate-limit
  retry and exhaustion, and truncation.
- Full suite: 369 passing, 1 skipped, 0 failures.
- `black --check` clean across the repo.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,16 +51,9 @@ VERTEX_EMBED_CONTENT_MODEL=gemini-embedding-001
 
 # Noise / Relevance Filter (Module B)
 
-# Dedicated cheap classification model for the Stage 2 gate. Decoupled from
-# CRE_LLM_CHAT_MODEL on purpose: Module B runs on every harvested commit and
-# must stay cheap. LiteLLM reads the provider key (GEMINI_API_KEY above)
-# automatically for gemini/* models.
 CRE_NOISE_FILTER_LLM_MODEL=gemini/gemini-2.5-flash-lite
-# Chunks per LLM request (cost/latency vs prompt-size tradeoff).
 CRE_NOISE_FILTER_BATCH_SIZE=10
-# Per-chunk character cap before sending to the LLM (longer chunks truncated).
 CRE_NOISE_FILTER_MAX_CHARS=1500
-# Minimum confidence for a KNOWLEDGE verdict to be enqueued (used from Week 4+).
 CRE_NOISE_FILTER_CONFIDENCE_THRESHOLD=0.8
 
 # Spreadsheet Auth
--- a/.env.example
+++ b/.env.example
@@ -49,6 +49,20 @@ GEMINI_API_KEY=your-gemini-api-key
 VERTEX_CHAT_MODEL=gemini-2.5-flash
 VERTEX_EMBED_CONTENT_MODEL=gemini-embedding-001
 
+# Noise / Relevance Filter (Module B)
+
+# Dedicated cheap classification model for the Stage 2 gate. Decoupled from
+# CRE_LLM_CHAT_MODEL on purpose: Module B runs on every harvested commit and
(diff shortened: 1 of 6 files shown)


northdpole (approved):
Good job on B-W3. The recall-first prompt and few-shots are clear, fail-safe parsing (UNCERTAIN on bad output / failed batches) is the right call for this gate, and the json_schema fallback only on capability gaps is a nice improvement over the older pattern. Fine to merge as the Week 3 slice — looking forward to queue + pipeline wiring in W4/W5.

.env.example
Pa04rth, line 1: remove unnecessary comments

manshusainishab: Done in the latest commit.

manshusainishab (commented):
updated

PR #949, opened 2026-06-25
feat(api): add GET /api/capabilities to expose myopencre feature flag

### LINKED ISSUE #948: MyOpenCRE feature invisible in all environments — missing GET /api/capabilities endpoint
### Summary
The MyOpenCRE feature is fully built and merged but does not appear in any environment. The nav link and the `/myopencre` route never render.

### Root cause
The frontend `useCapabilities` hook (`application/frontend/src/hooks/useCapabilities.ts`) fetches `GET /api/capabilities` expecting `{ "myopencre": <bool> }`. That route does not exist in `application/web/web_main.py`, so the request 404s. The hook's `.catch` then resolves `{ myopencre: false }` everywhere, which gates out:
- the MyOpenCRE nav link in `Header.tsx`
- the `/myopencre` route in `routes.tsx`

### Expected
`GET /api/capabilities` returns `200` with `{ "myopencre": <bool> }`, driven by a deployment flag. When the flag is on, the nav link and `/myopencre` route appear; when off, they don't — and the hook gets a clean `200 { "myopencre": false }` instead of a 404.

### Scope
- Add one backend route + one feature-flag function + tests.
- No frontend changes (the hook already consumes this contract).
- Do not touch the MyOpenCRE pages, parser, or import/export routes.

skypank-coder:
Adds the missing `GET /api/capabilities` endpoint that the frontend `useCapabilities` hook depends on. Without it the fetch 404s and the whole MyOpenCRE feature stays hidden in every environment.

### Changes
- New route `GET /api/capabilities` in `application/web/web_main.py` returning `{ "myopencre": <bool> }`, mirroring the existing `/rest/v1/config` thin-handler style.
- New flag `is_myopencre_enabled()` in `application/feature_flags.py`, reading `CRE_ENABLE_MYOPENCRE` via the existing `TRUE_VALUES` pattern. Defaults to off, so no existing deployment behavior changes.
- Tests in `application/tests/web_main_test.py`: key present + boolean, false-by-default, true-when-flag-set.

### Why a dedicated flag
MyOpenCRE download/preview is read-only and should be controllable independently of `CRE_ALLOW_IMPORT`. The new flag decouples its visibility from the import feature.

### Verification
- Frontend already consumes this contract (`useCapabilities.ts` → `Header.tsx`, `routes.tsx`); no FE change needed.
- Live-tested: flag off → `200 {"myopencre": false}` (clean 200, no more 404); flag on → `200 {"myopencre": true}`.
- New tests pass; black 24.4.2 and mypy clean on changed files.

Closes #948 

cc @PRAteek-singHWY  @Pa04rth  @northdpole

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,6 +31,16 @@ INSECURE_REQUESTS=false
 
 CRE_ENABLE_HEALTH=false
 
+# Expose MyOpenCRE in the UI (GET /api/capabilities → myopencre).
+# Off by default so production can roll out independently of import auth.
+
+CRE_ENABLE_MYOPENCRE=false
+
+# Expose Login/Logout in the UI (GET /api/capabilities → login).
+# Off by default until auth routes + User model land; enable on staging first.
+
+CRE_ENABLE_LOGIN=false
+
 # Embeddings
 
 NO_GEN_EMBEDDINGS=false
--- a/README.md
+++ b/README.md
@@ -289,7 +289,7 @@ Then edit `.env` and provide values appropriate for your environment.
 * Google Auth: `GOOGLE_CLIENT_ID`, `GOOGLE_CLIENT_SECRET`, `GOOGLE_SECRET_JSON`, `LOGIN_ALLOWED_DOMAINS`
 * GCP: `GCP_NATIVE`
 * Spreadsheet Auth: `OpenCRE_gspread_Auth`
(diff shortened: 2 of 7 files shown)


northdpole (approved):
Looks good. Extended this PR with `capabilities.login` / `CRE_ENABLE_LOGIN` (defaults off) so #943 can gate Login/Logout without a separate follow-up for the API contract. Closes the backend half of #964.

PR #937, opened 2026-06-18
week_3: Module C (The Librarian) — C.1 candidate retriever (in-memory + pgvector) + pipeline switch

PRAteek-singHWY:
# week_3: Module C (The Librarian) — C.1 semantic candidate retriever + pipeline switch

> **Stacked on #925 (Week 2).** This branch is based on `gsocmodule_C_week_2`; only the top commit is new. Until #922 and #925 merge, the diff shows the W1+W2 commits too — I will rebase onto `main` as they land, shrinking it to the Week 3 files only.

## Overview

Week 1 built the **judge** (golden dataset + scorer + harness). Week 2 built the parts that **must never guess** — the C.0 input boundary and the deterministic explicit-CRE fast path. This is the first part that **does** guess.

**The problem:** the explicit fast path only fires when a CRE id (`ddd-ddd`) is literally written in the chunk. Most chunks just describe a security concept in plain English, so for those C must *find* which of OpenCRE's CRE nodes the text is about. That is a search problem — and the first place in the pipeline that can be wrong.

**This PR's role:** build the **search step (C.1)**. Embed the section text, cosine-match it against the CRE-node vector hub, and return the top-K (default 20) rank-ordered candidates — the shortlist the W4 cross-encoder reranks and later weeks turn into a confident yes/no. Because Week 1 already built the judge, this first ML component gets a measured score from day one (recall@k, see Results).

**Scope:** 1 new module + benchmark + 1 new test, plus config/CLI/harness wiring. No frontend, no migration, no behaviour change to OpenCRE proper. (The pgvector schema migration lands in W8 — see "not here".)

## What changed

| Area | Files | Description |
|---|---|---|
| C.1 retriever | `candidate_retriever.py` *(new)* | Two interchangeable backends behind one `retrieve()` seam, both emitting the same RFC `RetrievalAudit` (`candidates[]` rank-ordered with `score_vector`; `reranked[]` empty until W4). **`CandidateRetriever`** — in-memory sklearn cosine over an injected `embed_fn` + immutable `CandidatePool`. **`PgVectorRetriever`** — pushes the cosine into Postgres via the `<=>` operator over an `embedding_vec` column (scored `1 - distance`), never loading the hub into RAM. `build_retriever(backend, ...)` is the factory. DI keeps the module import-light and hermetically testable (same pattern as W2's resolver). **Reuses** OpenCRE's embedding stack (`PromptHandler.get_text_embeddings`, `db.get_embeddings_by_doc_type`, `cosine_similarity`). |
[...]

PR #925, opened 2026-06-10
week_2: Module C (The Librarian) — C.0 input boundary: SectionValidator + ExplicitLinkResolver

PRAteek-singHWY:
# week_2: Module C (The Librarian) — C.0 deterministic input boundary: SectionValidator + ExplicitLinkResolver

> **Stacked on #922.** This branch is based on `gsocmodule_C_week_1`; only the top commit is new. I will rebase onto `main` as soon as #922 merges, which will shrink the diff to the Week 2 files only.

## Overview

This is the Week 2 deliverable for Module C (The Librarian): **C.0, the deterministic input boundary**. Before any retrieval or ML, every incoming chunk passes two deterministic stages:

1. **SectionValidator** — is this input well-formed and usable? Validates the row, adapts it into the internal `Section` the pipeline consumes, and rejects bad input with typed errors.
2. **ExplicitLinkResolver** — does the text already cite a CRE id (`ddd-ddd`)? If exactly one *known* id is cited, link it directly with no ML. Unknown or conflicting references never auto-link — they route to human review (fail-safe).

A naming note for reviewers: the plan documents called this component "SectionNormalizer", but the RFC (#734) assigns text normalization to **Module A** (harvest + normalize + chunk), with Module B's sanitizer on top. By the time text reaches Module C it is contractually clean, and re-cleaning it here would silently drift C from what A hashed and B classified. This component validates and adapts — it never transforms text — so it is named **SectionValidator**.

**Scope:** 4 new files + 1 updated (the eval harness). No frontend, no migrations, no DB access, no behaviour change to OpenCRE proper.

## What changed

| Area | Files | Description |
|---|---|---|
| C.0 validator | `section_validator.py` | Validates both upstream shapes — Module B's reduced `knowledge_queue` row and the full RFC `KnowledgeItem` envelope — into a frozen internal `Section`. Synthesizes the RFC identity fields from B's row (`chunk_id = chk:{repo}@{sha}:{path}`, `artifact_id = art:{repo}:{path}`, `source`, `locator`). Strips volatile audit metadata (`llm_reasoning`, filter stages, run timestamps) so downstream stages can never key decisions on it. Rejections are typed (`MalformedKnowledgeItemError`, `EmptyTextError`, `UnsupportedLanguageError`, `NotKnowledgeError`); raw Pydantic `ValidationError` never escapes the boundary. |
[...]

PR #922, opened 2026-06-09
week_1: Module C (The Librarian) — RFC contracts + eval harness + golden   dataset

PRAteek-singHWY:
# week_1: Module C (The Librarian) — contracts + eval harness + golden dataset

## Overview

This is the Week 1 deliverable for Module C (The Librarian). Following the OIE RFC's "build the test before the code" principle, this PR establishes the **data contracts and the evaluation harness for Module C — before any linking logic.** There is no retriever, no embeddings, and no decision logic here; those land in later weeks and will be measured against what this PR provides.

Module C consumes a `KnowledgeItem` from Module B and produces one of two outputs: a `LinkProposal` (auto-link to the OpenCRE graph) or a `ReviewItem` (route to human review in Module D). This PR makes those contracts concrete, guards them against drift from the RFC, and builds an evaluation harness plus a real 319-row dataset so that every future change to Module C is verifiable in CI.

**Scope:** 24 files. The only change to existing code is a one-line `pydantic` version pin. No frontend, no migrations, no pipeline wiring.

## What changed

| Area | Files | Description |
|---|---|---|
| Contracts | `schemas.py` | The three RFC envelopes as Pydantic v2 models — `KnowledgeItem` (input), `LinkProposal` / `ReviewItem` (output) — plus shared sub-models. The RFC's conditional rules are enforced as validators. |
| Drift guard | `_rfc_schemas/*.json` (6) + `schemas_test.py` | The canonical RFC schemas, vendored and pinned to a specific upstream commit. A test validates every model against them, so CI fails if our contracts diverge from the RFC. |
| Config | `config_loader.py` | The seven `CRE_LIBRARIAN_*` settings (thresholds, model name, top-k) loaded into one frozen config. |
| Input interface | `knowledge_source.py` | An abstract `KnowledgeSource` interface with a fixture-backed stub. The real database-backed reader arrives later and yields the same shape. |
| Eval harness | `evaluate_librarian.py` | Loads the dataset, applies the hub-firewall and scorer, and prints per-slice results. This is the baseline that every later week is measured against. |
| Hub-firewall | `hub_firewall.py` | Strips each test row's own text out of the CRE hub before scoring, so retrieval cannot echo the answer back and inflate accuracy. On by default. |
| Scoring | `scoring.py` | The multi-link correctness rule: a prediction is correct only if Jaccard(expected, predicted) ≥ 0.5 and the top-1 prediction is in the expected set. |
[...]

PR #957, opened 2026-07-03
week_4: Module C (The Librarian) — C.2 cross-encoder reranker

PRAteek-singHWY:
Hi @northdpole — Week 4 of Module C. This one adds the cross-encoder that re-reads the shortlist C.1 produced.

> **Stacked on #937 (Week 3).** Based on `gsocmodule_C_week_3`, so until #922 → #925 → #937 land the diff shows the earlier commits too — I'll rebase onto `main` as they merge, shrinking it to just the W4 files.

## The problem it fixes

Week 3's search is fast but rough. The bi-encoder fingerprints the section and each CRE *separately*, then compares them — great for narrowing hundreds of CREs down to 20, but it never actually reads a section and a CRE *together*, so the exact ordering inside those 20 is unreliable. The real answer might be sitting at #7, not #1.

## What this does

Takes those 20 candidates and reads each one *side-by-side with the section* as a single combined input, and scores "do these two actually match?" — then re-sorts the 20 and keeps the best 5. It fills the `reranked[]` slot we deliberately left empty in Week 3, and leaves `candidates[]` untouched so the pre-rerank shortlist stays auditable.

Two model kinds, one line: the **bi-encoder (W3)** fingerprints each thing alone — fast, whole-hub, rough. The **cross-encoder (W4)** reads the pair together — slow, so we only run it on the 20, but much more accurate. W3 skims 20 résumés to make a shortlist; W4 interviews those 20 one-on-one to pick the top 5.

## What changed

- **`cross_encoder.py`** *(new)* — `CrossEncoderReranker` over an injected `score_fn`, the same DI pattern as C.1's `embed_fn`, so the module never imports torch and stays hermetically testable. `build_cross_encoder_score_fn` lazily loads the pinned `ms-marco-MiniLM-L-6-v2`. Stable sort so ties keep C.1's cosine order; typed errors; a `RERANKER_NAME` audit tag.
- **`cross_encoder_test.py`** *(new)* — 9 hermetic tests (re-ordering by cross-encoder score, top-N truncation, tie stability, audit shape, and the missing-text / score-count failure modes).
- **`db.py`** — `get_embedding_contents_by_doc_type`: the `{id → embeddings_content}` pair text the reranker scores against (mirrors `get_embeddings_by_doc_type`).
- **`cre_main.py`** — `run_librarian` now reranks C.1's audit and logs the reranked top-N per section.
- **`evaluate_librarian.py`** — v1 full pipeline; reports the live rerank top-1 alongside recall@k. Offline path unchanged.
- **`requirements.txt`** — add `sentence-transformers`; `__init__.py` scope note now covers C.2.

## Results (positive slice, live)

```
retrieval recall@20 (C.1): any-hit 285/292 (98%), all-hit 274/292 (94%)
rerank top-1     (C.2): 220/292 (75%)   (target >= 0.80)
```

Measured live against a populated `standards_cache.sqlite` (428 CRE embeddings, `gemini/gemini-embedding-001`, dim 3072), hub-firewall ON, `top_n_rerank=5`.

**Reading it:** the retriever puts the right CRE in the top-20 **98%** of the time, so the shortlist the reranker sees is almost always complete. The cross-encoder then lands the correct CRE at **#1 for 75%** of rows — solid, but **5 points under the 0.80 target**. The gap is the honest state at W4: the reranker sometimes demotes a correct near-tie. Calibration (W5) + the threshold experiment (W7) are the levers to close it; the target is a W6/W8 gate, not a W4 merge blocker. The reranker logic itself is fully covered offline by the hermetic tests above.

## Not here (later weeks)

- Calibration + the confident yes/no decision (C.3–C.4, W5–W6).
- Graph writes / wiring into the worker (W8) — still dry-run only; the CLI stays opt-in and only costs the embedding API on manual runs.

### DIFF
--- a/application/cmd/cre_main.py
+++ b/application/cmd/cre_main.py
@@ -1073,10 +1073,10 @@ def run_librarian(
 
     For each knowledge-queue section: try the deterministic explicit-CRE fast
     path (C.0.5); on no/ambiguous reference, run the semantic retriever (C.1)
-    and log the top-K candidate shortlist. The cross-encoder rerank (C.2, W4),
-    decision/threshold routing (C.3-C.4, W5) and graph writes (W8) are not
-    built yet, so this is dry-run only: it never writes a link. ``--run_librarian``
-    without writes behaves identically and warns.
+    and rerank its top-K shortlist with the cross-encoder (C.2, W4), then log
+    the reranked candidates. Decision/threshold routing (C.3-C.4, W5) and graph
+    writes (W8) are not built yet, so this is dry-run only: it never writes a
+    link. ``--run_librarian`` without writes behaves identically and warns.
 
     Ops note: this is opt-in CLI only — it is not on the ``Procfile`` and is
     wired into neither the web app nor the background worker (that lands W8).
@@ -1172,9 +1172,19 @@ def run_librarian(
             logger.info("[explicit] %s -> %s", section.chunk_id, resolution.cre_ids[0])
             continue
 
+        try:
+            audit = retriever.retrieve(section.text)
+            audit = reranker.rerank(section.text, audit)
+        except Exception as exc:  # noqa: BLE001 - one bad section must not abort the batch
+            rejected += 1
+            logger.warning(
+                "[semantic] %s skipped: retrieval/rerank failed: %s",
+                section.chunk_id,
+                exc,
+            )
+            continue
+
         semantic += 1
-        audit = retriever.retrieve(section.text)
-        audit = reranker.rerank(section.text, audit)
         top = ", ".join(
             f"{c.cre_id}:{c.score_rerank:.3f}" for c in audit.reranked
         )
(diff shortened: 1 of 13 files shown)


northdpole (approved):
Strong W4 — cross-encoder reranker integrates cleanly with C.1 shortlist, dry-run CLI path is well-scoped (no writes until W8), and config/tests cover the new surface. Ready to merge after #925 and #937 land.

PRAteek-singHWY:
> Strong W4 — cross-encoder reranker integrates cleanly with C.1 shortlist, dry-run CLI path is well-scoped (no writes until W8), and config/tests cover the new surface. Ready to merge after #925 and #937 land.

Thank you @northdpole , will resolve these conflicts altogether once #937 gets merged.
Okay this got merged, on it will notify you once these are resolved.

northdpole (approved):
Re-approved after W4-only rebase onto main (#937 merged). C.2 cross-encoder + live eval id-space fix look good.

PR #959, opened 2026-07-07
Remove unused graph-visualization dependencies

### LINKED ISSUE #958: Drop unused react-digraph, react-flow, and 3 other dead deps
package.json lists 5 graph/visualization libraries that aren't imported anywhere in application/frontend/src
package.json is the file that declares every package your project depends on. Having a package listed there means it gets downloaded during yarn install  but it doesn't mean any code actually uses it. The claim is: 5 specific packages are declared but dead  nothing in the actual source code (application/frontend/src) ever writes import ... from '<package>' for them.

The 5 packages, named explicitly:
Package	                            Status
react-digraph	                   Unused
react-flow	                   Unused (see note below)
react-force-graph-2d	   Unused
react-d3-library	          Unused
d3-dag	                          Unused

"react-flow (superseded by react-flow-renderer, which is used)"
This is a parenthetical clarifying why react-flow specifically is dead, so a reviewer doesn't think it's a typo. Here's the history:

The graph-layout library "React Flow" was published under the npm name react-flow-renderer for versions 1 through 9.
Starting at version 10, the maintainers renamed the package to just react-flow.

This repo has both names installed  but the actual code only imports from react-flow-renderer (the older name). So react-flow isn't a different library sitting unused; it's literally the same project under its other name, installed a second time for nothing.
"These add install time and unnecessary attack surface with no functional benefit"
Three separate costs are being named here:

Install time every one of these 5 packages (plus their own sub-dependencies) gets downloaded on every yarn install / CI run, even though the build never touches them. Pure wasted download/disk time.

Attack surface "attack surface" means the total amount of code that could contain a security vulnerability. Every installed package is something a CVE (security vulnerability disclosure) could show up in  and someone would then have to evaluate/patch it, even though the vulnerable code is never actually executed here. Fewer unused packages = fewer things to worry about.

No functional benefit removing them changes nothing about how the app behaves, since nothing calls into them today. This is the reassurance that the fix is low-risk: it's pure subtraction, not a behavior change.

In one sentence: "We're carrying 5 packages that cost install time and security-review overhead but do nothing, and here's the proof."

DevPatils:
Removes 5 graph-visualization dependencies from `package.json` that are
declared but never imported anywhere in `application/frontend/src`:

- `react-digraph`
- `react-flow` (superseded by `react-flow-renderer`, which *is* used —
  same library, renamed at v10; both were installed, only the old name
  is referenced in code)
- `react-force-graph-2d`
- `react-d3-library`
- `d3-dag`

Fixes #958

## Why

These packages add install time and unnecessary dependency/security
surface with zero functional benefit — nothing in the codebase imports
them. Confirmed via repo-wide search across `application/frontend/src`
before removal.

## Changes

- `package.json` — removed the 5 unused entries
- `yarn.lock` — regenerated via `yarn install`
- `application/frontend/www/bundle.js` / `bundle.js.LICENSE.txt` —
  rebuilt production bundle reflecting the dependency change (this repo
  deploys the committed bundle directly, no separate CI build step)

## Test plan

- [x] `yarn install` completes cleanly with no missing-peer errors
      introduced by the removal
- [x] `yarn build` (`webpack --config webpack.prod.js`) succeeds
- [x] Full TypeScript type-check (via `awesome-typescript-loader` in
      the build) passes with zero errors, confirming no file anywhere
      referenced the removed packages
- [x] Bundle size unchanged in behavior, slightly reduced in size
      (3.45 MiB → 3.3 MiB)


DevPatils:
@Pa04rth Please review my changes :)

northdpole:
Thanks @DevPatils — confirmed the removed packages aren't imported anywhere, security-relevant deps (dompurify, axios, OAuth) are unchanged, and this is just dead lockfile/bundle cleanup. Merging.

northdpole (approved):
LGTM — unused deps only, no security/runtime impact.

PR #876, opened 2026-04-09
RFC: User Authentication and Online MyOpenCRE Mapping

skypank-coder:
## Summary

This RFC proposes the design for user authentication and 
personal MyOpenCRE views (issue #586).

Reviewed and approved by @northdpole — opening for 
community review as suggested.

## What this RFC covers

- OAuth login via Google (no password storage)
- Persistent user identity using `google_sub` as anchor
- User-owned standard and section storage with no 
  data duplication against the shared CRE graph
- Automated mapping execution pipeline (bi-encoder + 
  cross-encoder) with 98% accuracy checkpoint
- Upload endpoint wiring under `/rest/v2/`
- Result retrieval with ownership enforcement

## Implementation order

5 TODO steps with explicit code changes, tests, 
pass/fail checkpoints, and rollback behaviour for each.

## Related

- Closes part of #586
- Related to #375 (privacy note included)
- New endpoints under `/rest/v2/` — no breaking changes 
  to existing `/rest/v1/` surface

## Reviewers

@northdpole @parth-sohaney @robvanderveer

Issue #586, opened 2025-02-22
Add support for users


northdpole:
OpenCRE could become more useful if users could make an account and select which resources they want to be working with.

This would allow users to not only see and export subgraphs of information but most importantly map their resources to sub-graphs without the noise.

This feature set contains the following sub-features:

* A user should be able to login with an OIDC provider (e.g. google, github, linkedin etc)
* A user should have the ability to select which resources they want to see in their account

Pa04rth:
Sir what are your requirements after login -
1- User must go to a new page or same home page 
2- which db you wanna use {I may use Mongodb} for user credentials.

I am willing to work on this issue ,Plz assign me this issue.

How I will solve the above issue -
1- I will use Auth0 for handling the signup and login mechanism 
2- using the above we can easily integrate the OIDC providers 
3- Creating a filter to filter out the resources

Pa04rth:
Please assign me this issue @northdpole

northdpole:
> Sir what are your requirements after login -
> 1- User must go to a new page or same home page
> 2- which db you wanna use {I may use Mongodb} for user credentials.
This should work seamlessly with the platform, we currently use postgresql as our database and sqlite for development all exposed via a sqlalchemy ORM.
A user should be able to filter the resources they want to see after logging in. 
e.g. not everyone cares about zap rules or asvs or samm or any other resource so they should be able to filter them out if they wish.
> 
> I am willing to work on this issue ,Plz assign me this issue.
> 
done
> How I will solve the above issue -
> 1- I will use Auth0 for handling the signup and login mechanism
> 2- using the above we can easily integrate the OIDC providers
> 3- Creating a filter to filter out the resources
let's start with the login first, let's create the login that can be turned off for local development and then we can work on the filtering.

abhishekblue:
I'd like to work on this. Can you assign me to this issue?

sd2604:
Hi I have noticed that this issue is inactive for a while, If its okay could you pls assign this to me if its still available?

skypank-coder:
The core per-user resource selection loop is complete and merged:

- User + resource-selection persistence (#980)
- `GET`/`PUT /rest/v1/user/resources` API (#981)
- Server-side filtering of `/rest/v1/standards` with `?all=true` bypass (#1001)
- Resource-selection UI in MyOpenCRE (#1005)
- Concurrency-safe selection write + OpenCRE-always-included invariant (#1008)

[...]

PR #973, opened 2026-07-09
feat(db): persist users + resource selection — Part of #586

skypank-coder:
## What & why
First increment of #586 ("Add support for users"): login-first user
persistence. OIDC login existed but nothing was persisted — no User table,
nowhere to hang per-user data. This adds that foundation so per-user resource
filtering (PR2–PR4) can be built on it. Aligns with auth RFC #876 (TODO 1/2).

## Changes
- **`users`** table anchored on the immutable OIDC `sub` (`google_sub`); email
  and display_name refreshed on each login.
- **`user_resource_selection`** table: one row per selected standard,
  `UNIQUE(user_id, standard_name)`, `FK → users.id ON DELETE CASCADE`.
- **Login callback** upserts the User and sets `session['user_id']` — gated on
  `CRE_ENABLE_LOGIN`, wrapped so it never blocks login. Existing session keys
  and the anonymous 401 are unchanged (intentionally narrower than RFC #876's
  `login_required` rewrite).
- **`Node_collection`** methods: `upsert_user`, `get_user_by_sub`,
  `get_user_resource_selection`, `set_user_resource_selection`. These back the
  `/rest/v1/user` resource endpoints coming in PR2.
- **Migrations:** a dedicated merge revision collapses the two pre-existing
  open heads (`ab12cd34ef56`, `9b1c2d3e4f50`), then a revision creates the two
  tables — keeping the head single-parent so `flask db downgrade` is
  unambiguous.

## Testing
- 12 new tests: idempotent upsert (no dup row), selection round-trip / replace /
  dedup / per-user isolation, unique constraint, cascade delete, callback upsert,
  and the flag-off no-op path.
- Verified on **real Postgres** (prod parity), not just SQLite: single head,
  `ON DELETE CASCADE` enforced at the DB level (rows 1 → 0), both unique
  constraints created cleanly, migration `upgrade → downgrade → upgrade`
  deterministic, alembic revision guardrail green. 12/12 pass on Postgres.
- `black` clean; no new mypy errors vs. baseline.

## Scope boundaries
No new endpoints, no server-side filtering, no frontend, no `login_required`
rewrite — those are the follow-up PRs in #586.

## Unrelated pre-existing issue (out of scope, flagging for visibility)
`flask db upgrade` from a **blank** DB fails on Postgres before reaching my
revisions: migrations `3c65127871a6` / `7bf4eac76958` reuse the constraint name
`uq_pair` (authored via SQLite `batch_alter_table`), which collides on a
from-base replay. Postgres rolls the whole thing back cleanly (transactional
DDL). This predates this branch and doesn't affect the prod path (apply-onto-
existing-schema), so it's out of scope here — but worth a separate fix, since
anyone bootstrapping a fresh Postgres DB will hit it.


migrations/versions/f0e1d2c3b4a5_merge_heads_before_users.py
```diff
@@ -0,0 +1,27 @@
...
+Create Date: 2026-07-08
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "f0e1d2c3b4a5"
+down_revision = ("ab12cd34ef56", "9b1c2d3e4f50")
```
northdpole, line 17: Good call merging the two embedding heads before adding users — keeps downgrade unambiguous and matches prod's two-row `alembic_version` state. No concerns on the no-op merge pattern.

migrations/versions/a1b2c3d4e5f6_add_users_and_resource_selection.py
```diff
@@ -0,0 +1,50 @@
...
+        sa.Column("last_seen_at", sa.DateTime(), nullable=False),
+        sa.UniqueConstraint("google_sub", name="uq_users_google_sub"),
+    )
+    op.create_table(
+        "user_resource_selection",
+        sa.Column("id", sa.String(), primary_key=True),
+        sa.Column("user_id", sa.String(), nullable=False),
+        sa.Column("standard_name", sa.String(), nullable=False),
+        sa.Column("created_at", sa.DateTime(), nullable=False),
+        sa.ForeignKeyConstraint(
```
northdpole, line 36: Clean additive migration. `ON DELETE CASCADE` on `user_id` matches the model and tests.

application/database/db.py
```diff
@@ -254,6 +254,49 @@ class StagedChangeSet(BaseModel):  # type: ignore
...
+class User(BaseModel):  # type: ignore
+    """Persisted account identity for OIDC login (issue #586, RFC #876 TODO 1).
+
+    ``google_sub`` is the immutable OIDC ``sub`` claim (the same value stored in
+    ``session['google_id']``); email can change, so identity is anchored on the sub.
+    """
+
+    __tablename__ = "users"
+    id = sqla.Column(sqla.String, primary_key=True, default=generate_uuid)
+    google_sub = sqla.Column(sqla.String, nullable=False, unique=True)
```
northdpole: Nit: `google_sub` has both `unique=True` on the column and `UniqueConstraint(..., name="uq_users_google_sub")` in `__table_args__`. The Alembic migration only creates the named constraint, so `sqla.create_all()` in tests may diverge (double unique). Suggest dropping the column-level `unique=True` and keeping only the named constraint.

application/database/db.py
```diff
@@ -992,6 +1035,76 @@ def with_graph(self) -> "Node_collection":
...
+    ) -> User:
+        """Create or refresh the User row for an OIDC ``sub`` (issue #586).
+
+        Idempotent on ``google_sub``: a repeat login updates the existing row's
+        profile fields and ``last_seen_at`` instead of inserting a duplicate.
+        """
+        from datetime import datetime, timezone
+
+        now = datetime.now(timezone.utc)
+        user = self.session.query(User).filter(User.google_sub == google_sub).first()
```
northdpole: Minor robustness: check-then-insert can race on concurrent logins for the same `google_sub`. Second `commit()` would raise `IntegrityError` and leave the session dirty. Worth mirroring the `add_node` pattern: catch `IntegrityError`, `rollback()`, re-query by `google_sub`, update profile fields, return. Low urgency for PR1 but cheap to add now.

application/web/web_main.py
```diff
@@ -1161,6 +1161,19 @@ def callback():
             401,
             description=f"You need an account with one of the following providers to access this functionality {allowed_domains}",
         )
+
+    # Persist the account when login is enabled; the session keeps working
+    # unchanged if this no-ops (flag off) or fails.
+    if is_login_enabled():
+        try:
+            user = db.Node_collection().upsert_user(
```
northdpole: Like the flag gate + broad `try/except` — login must not break if persistence fails. Two small notes:
1. `id_info.get("sub")` can be `None`; consider bailing early (log + skip upsert) rather than hitting a NOT NULL violation.
2. `Node_collection()` initializes `NEO_DB.instance()` unless `NO_LOAD_GRAPH_DB=1`. Probably fine (singleton), but a lighter login path could avoid pulling Neo4j into the auth callback. Non-blocking for PR1.

application/tests/web_main_test.py
```diff
@@ -1151,6 +1151,68 @@ def test_gap_analysis_weak_links_response(self, db_mock) -> None:
             self.assertEqual(200, response.status_code)
             self.assertEqual(expected, json.loads(response.data))
 
+    @patch("application.web.web_main.id_token")
+    @patch("application.web.web_main.CREFlow")
+    def test_callback_upserts_user_and_sets_session(
```
northdpole, line 1156: Good coverage for flag-on persist and flag-off no-op. Small addition: assert `session['user_id']` equals the persisted `User.id` in the flag-on test — catches session/DB mismatch before PR2 wires `/rest/v1/user` resource endpoints.

application/tests/user_model_test.py
```diff
@@ -0,0 +1,164 @@
+"""Tests for user persistence and per-user resource selection (issue #586, RFC #876 TODO 1/2)."""
```
northdpole, line 1: Solid SQL-layer coverage: idempotent upsert, selection round-trip/replace/dedup/isolation, unique constraint, cascade delete. Postgres verification in the PR description is the right bar for prod parity.

northdpole (approved):
## Alembic / deploy

Verified migration graph locally; prod `alembic_version` confirmed — both embedding heads (`ab12cd34ef56`, `9b1c2d3e4f50`) present. This PR correctly merges them (no-op) then adds two new tables. Deploy path is additive only.

## Prod impact

Low. Runtime change is flag-gated (`CRE_ENABLE_LOGIN`) and failure-safe on callback. No user-facing behavior in PR1. LGTM on scope for #586 / RFC #876 TODO 1.

**Approved.** A few non-blocking nits inline — address now or track for a follow-up.

PR #974, opened 2026-07-09
week_5: Module C (The Librarian) — C.3 confidence calibration (temperature scaling)

PRAteek-singHWY:
Hi @northdpole — Week 5 of Module C. This one turns the cross-encoder's raw scores into an **honest probability** (temperature scaling), so Week 6 can threshold "auto-link vs. send-to-human" on a number that actually means what it says.

> **Based on latest `main`** — rebased/merged onto `main` after Week 1–4 (#922 → #925 → #937 → #957) and after the pgvector migration (#979), so the diff is a clean **Week-5-only** surface.

## Overview

Week 3 built the **search step (C.1)** — a bi-encoder that cosine-ranks the whole CRE hub down to a top-20 shortlist. Week 4 built the **rerank step (C.2)** — a cross-encoder that reads each `(section, candidate-CRE)` pair together and re-sorts the shortlist by a raw relevance logit. Those logits are great for *ordering* but are **not** confidence: a raw `+1.5` does not mean "82% sure this is the right CRE," and cross-encoders are systematically over-confident.

**The problem:** Week 6's decision — auto-link vs. send to a human — is a **threshold on confidence** ("link if ≥ 90%"). That is only safe if 90% really means 90%. A raw score cannot be thresholded honestly.

**This PR's role:** build the **calibration step (C.3)** — temperature scaling. Learn a single scalar **temperature `T`** on the golden set and map the reranked shortlist to a real probability `p = softmax(logits / T)`; the top-1 candidate's share of that probability mass is the confidence — "how likely is the top candidate the correct CRE?" Prove it honest with **Expected Calibration Error < 0.10**. Temperature only flattens/sharpens the distribution — it **never changes the ranking** — so calibration can't hurt C.1/C.2 recall; it only makes the confidence trustworthy.

**One design note worth surfacing** (the golden set caught it): calibrating the *single absolute top-1 logit* with `sigmoid(z / T)` does **not** work — a cross-encoder's absolute logit has no fixed zero point (its 50/50 boundary isn't at `z=0`), and one temperature can only squash toward 0.5, never *shift* the boundary, so ECE stays stuck at ~0.27. Calibrating the **softmax over the whole shortlist** is the fix (and the textbook Guo et al. temperature scaling): the candidates' *relative* logits are what a cross-encoder's scores actually mean, and their softmax is a genuine distribution over "which candidate is right." It is near-calibrated even before fitting (ECE 0.053 at `T=1`), and `T` polishes it to **0.046**. (`CALIBRATOR_NAME = "temperature-scaling/0.2.0"` records this over the rejected single-logit `0.1.0`.)

**Scope:** 1 new package (1 module + 1 test) + harness wiring for the ECE gate. `UpdateDetection` (the other W5 schema deliverable) **already exists** in `schemas.py` — verified, nothing to add. No frontend, no behaviour change to OpenCRE proper.

## What changed

| Area | Files | Description |
|---|---|---|
| C.3 calibration | `calibration/temperature.py` *(new)*, `calibration/__init__.py` *(new)* | `TemperatureScaler(T)` maps a reranked shortlist to a calibrated distribution (`.probabilities` = `softmax(logits/T)`) and a top-1 `.confidence` (the mass Week 6 thresholds on); `fit_temperature` learns the single scalar `T` by minimising NLL (`scipy.optimize.minimize_scalar`, bounded), guarding single-outcome data (`DegenerateLabelsError`); `expected_calibration_error` is the 10-bin ECE gate; `negative_log_likelihood` is the fit objective (exposed for testing/reporting). Model-free (numpy + `scipy.special.softmax`) so it stays import-light and hermetically testable — mirrors the C.1 `embed_fn` / C.2 `score_fn` seams. `CALIBRATOR_NAME` audit tag (mirrors `RETRIEVER_NAME`/`RERANKER_NAME`). |
| Eval harness | `evaluate_librarian.py` | `report_calibration` builds a `(shortlist, label)` set from the live C.1→C.2 pipeline over the **positive + hard_negative** slices (each row's reranked logits; label `1` iff its top-1 is an expected CRE — hard_negatives supply the `0` outcome), fits `T`, and reports **ECE at T=1 vs. the fitted T** with the **< 0.10** gate. `_build_live_pipeline` extracted so recall and calibration share one hub + model load. Offline path unchanged (no live number is faked). |
| Docs | `__init__.py` | Scope note now covers C.3. |
| Tests | `temperature_test.py` *(new)* | 21 hermetic tests: softmax top-1 confidence, temperature flatten/sharpen, single-candidate edge, NLL hand-check, fit **recovers a known `T`** and **reduces NLL vs. T=1**, ECE on perfectly-/mis-calibrated data + a hand-checked two-bin value, peaked-vs-flat shortlist confidence, and every guard (single-outcome, non-binary, length mismatch, empty, `n_bins < 1`, non-positive/non-finite `T`). |

## How the pieces connect

```mermaid
flowchart TB
    audit["RetrievalAudit from C.2 (W4)<br/>reranked[] shortlist with score_rerank logits"]
    subgraph FIT["fit once on the golden set"]
        g["(shortlist, is-top1-correct) pairs<br/>positive + hard_negative slices"]
        nll["fit_temperature: argmin NLL(T)"]
        Tstar["T-star (one scalar)"]
        ece["expected_calibration_error<br/>gate: ECE below 0.10"]
        g --> nll --> Tstar --> ece
    end
    subgraph APPLY["apply per shortlist"]
        s["reranked logits"]
        p["confidence = top-1 mass of<br/>softmax(logits / T-star)"]
        s --> p
    end
    audit --> g
    audit --> s
    Tstar -.-> p
    p --> dec["W6 decision engine<br/>confidence at or above tau: LinkProposal, else ReviewItem"]
```

## Results

```
# offline (CI default) — hermetic, no key/DB/model
113 librarian tests passing (92 from W1–W4 + 21 new; 1 skipped)
explicit slice (C.0.5 resolver): 5/5 — gate 100%: PASS
calibration (C.3): wired; ECE gate needs --use_live_embeddings

# live (local, against a populated standards_cache.sqlite migrated to embedding_vec:
# 428 CRE embeddings, gemini/gemini-embedding-001 dim 3072, ms-marco-MiniLM-L-6-v2, hub-firewall ON)
retrieval recall@20 (C.1): any-hit 285/292 (98%), all-hit 274/292 (94%)
rerank top-1     (C.2): 220/292 (75%)
calibration (C.3, 304 rows): fitted T=1.105; ECE 0.053 (raw, T=1) -> 0.046 (calibrated); gate ECE<0.10: PASS
```

Reading it: the reranked top-1 is correct 75% of the time, and after calibration the model's stated confidence tracks that — **mean confidence 0.767 vs. actual accuracy 0.757**, ECE **0.046 < 0.10**. So a "90%" from C.3 can be trusted as ~90%, which is exactly what the W6 auto-link threshold needs. (Calibration doesn't touch the 75% top-1 itself — that's the reranker's job, with W6/W7 the levers to lift it toward 0.80.)

## What is intentionally not here

- **Decision / threshold routing (C.4, W6)** — turning the calibrated confidence into a `LinkProposal` (≥ τ) vs. a `ReviewItem`, and writing it into `ProposedLink.confidence`. This PR produces the honest number; W6 thresholds it.
- **Persisting the fitted `T`** for the live decision path (loaded by the W6 engine).
- **Graph writes / worker wiring (W8)** — the pipeline stays dry-run; nothing here costs the embedding API outside a manual `--use_live_embeddings` run.

## How to verify locally

```bash
# all librarian tests (calibration is hermetic — no key, DB, or model)
python3 -m unittest discover -s application/tests/librarian -p '*_test.py' -t .
# or just the new ones:
python3 -m unittest application.tests.librarian.temperature_test

# offline harness: C.0 pass rate + explicit gate + calibration wired
python3 scripts/evaluate_librarian.py --dataset application/tests/librarian/fixtures/golden_dataset.json

# live ECE gate (needs a populated cache DB + an embedding-capable LLM)
# NOTE: after the pgvector migration (#979) a legacy SQLite cache must first be
# rewritten to the embedding_vec store, or the ORM will refuse it:
python3 scripts/rewrite_sqlite_embeddings_to_vec.py --db standards_cache.sqlite
python3 scripts/evaluate_librarian.py \
    --dataset application/tests/librarian/fixtures/golden_dataset.json \
    --use_live_embeddings --cache_file standards_cache.sqlite
```


scripts/evaluate_librarian.py
```diff
@@ -193,6 +208,62 @@ def _to_ext(mapping):
...
+        expected = set(row.expected.cre_ids or [])
+        logit_sets.append([float(c.score_rerank) for c in reranked])
+        labels.append(1.0 if reranked[0].cre_id in expected else 0.0)
+
+    if len(set(labels)) < 2:
+        print(
+            "calibration (C.3): need both outcomes in the selection (positive + "
+            "hard_negative slices) to fit temperature; skipped"
+        )
+        return 0
```
northdpole: **High — calibration skip reports success**

When the live C.3 calibration set is degenerate (all top-1 labels 0 or all 1, empty after dropping empty shortlists, or `--slice` with only one class), `report_calibration` prints "skipped" and **returns 0**. `main` then exits successfully even though the ECE < 0.10 gate never ran.

With `--use_live_embeddings`, a skipped gate should fail the run (non-zero exit), e.g. `return 1`, so CI/harness cannot greenwash a missing calibration. Only return 0 when calibration actually ran and ECE passed.

```python
if len(set(labels)) < 2:
    print(
        "calibration (C.3): need both outcomes in the selection (positive + "
        "hard_negative slices) to fit temperature; FAILED (gate did not run)"
    )
    return 1
```

Please fix before merge.

northdpole (changes requested):
## Maintainer review — Module C Week 5 (#974)

Temperature scaling (softmax-over-shortlist + NLL fit + ECE) looks correct and the hermetic tests are strong. One **gate integrity** issue before merge.

**Please fix the inline comment** (calibration skip must not exit 0 when `--use_live_embeddings` is on). After that, this is merge-ready as the base of the Module C stack (#974 → #990 → #991).

scripts/evaluate_librarian.py
PRAteek-singHWY: Fixed in `62abf2d2`. A degenerate calibration set now returns 1 instead of 0, so `--use_live_embeddings` cannot exit 0 without the ECE gate having run. The message also prints the row and class counts, so a failure says why the set was degenerate rather than just that it was.

Verified the hermetic path is unaffected: `evaluate_librarian.py --dataset application/tests/librarian/fixtures/golden_dataset.json` still exits 0, since calibration is skipped upstream of this check when there are no live CRE vectors.

PRAteek-singHWY:
Pushed `2bbc76e5` on top of the gate fix. Bot-review follow-ups on this PR:

1. "Live pipeline is built twice per run" was already stale: `main` builds the retriever and reranker once and passes them into both reports. What was still real is that the *shortlists* were recomputed, so `report_retrieval_recall` and `report_calibration` each re-ran retrieve and rerank over the positive slice. `live_audits()` now retrieves and reranks each row once, keyed by golden row id, and both reports read the same audits. Rows without an audit no longer count toward a report's denominator, so the printed fractions cannot divide by unscored rows.

2. `temperature.py`: `_softmax_top` duplicated the empty-shortlist guard and the softmax already in `TemperatureScaler.probabilities`. Both now route through one `_softmax_at` helper, and `confidence()` derives from `probabilities()`, so the two cannot disagree.

3. `temperature.py` is now clean under the `--strict` mypy the coding guidelines ask for: annotated `_paired`'s return type and the `bounds` tuple, re-asserted the array type over untyped scipy, and hoisted the label conversion out of the fit objective (it was re-validated on every optimiser iteration).

113 librarian tests pass and the hermetic harness run still exits 0. No metric moved.

scripts/evaluate_librarian.py
```diff
@@ -164,15 +157,62 @@ def _to_ext(mapping):
...
+    Every live report wants the same thing per row: the reranked shortlist. The
+    cross-encoder pass is the expensive step (one inference per candidate pair),
+    and the positive slice is read by more than one report, so computing the
+    audits here means a live run pays for exactly one retrieve + rerank per row
+    regardless of how many reports consume it.
+    """
+    return {
+        row.id: reranker.rerank(row.input.text, retriever.retrieve(row.input.text))
+        for row in rows
+    }
```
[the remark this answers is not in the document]
PRAteek-singHWY: Good catch, this was a real trap the shared-audit refactor opened. Fixed in `f352dfcc`.

Before the refactor each report recomputed per row, so a duplicate id was harmless. Keying the shared audits by `row.id` turned it into a silently wrong number: the surviving audit would be reused for the earlier row and scored against the wrong text. `load_dataset` now fails fast on duplicate ids and names the offenders, since `GoldenDatasetRow` only requires an id to be non-empty. The committed dataset has 319 distinct ids so nothing changes today. Added a test that a forced collision raises.

northdpole (approved):
## Re-review — gate fix verified

The Jul 31 blocker is addressed: `report_calibration` now returns **1** on a degenerate set (`len(set(labels)) < 2`) with an explicit FAILED message, so `--use_live_embeddings` cannot exit 0 without the ECE gate having run. ECE FAIL still returns 1 as before.

Duplicate-id fail-fast in `load_dataset` also looks correct for the shared-audits refactor.

CI green + mergeable. Approving and merging as the Module C stack base (#974 → #990 → #991).

PR #979, opened 2026-07-13
feat: pgvector embedding_vec migration + chat similarity cutover

northdpole:
## Summary
- Merge dual Alembic heads (`9b1c2d3e4f50`, `ab12cd34ef56`) in one revision `c7d8e9f0a1b2` that enables `vector`, adds `embeddings.embedding_vec`, **backfills from CSV, then drops the legacy CSV `embeddings` column** (hard-fails on mixed dims / incomplete backfill; no LLM re-embed on Heroku).
- **`embedding_vec` is the sole store** — `add_embedding` and readers write/parse vec literals only (Postgres `vector(N)`; SQLite/CI Text literals).
- Chat/import similarity prefers Postgres `<=>` when ready; otherwise sklearn over parsed `embedding_vec`. Requesting pgvector on SQLite/`can_use_pgvector_similarity()==False` **SystemExit**s with a clear message (no silent fallback).
- Module C / Librarian: `in_memory` loads vec literals into RAM; `CRE_LIBRARIAN_RETRIEVER_BACKEND=pgvector` requires Postgres + column.
- Helpers + SQLite cache rewrite / sync scripts updated for vec-only.

Closes #977.

## Test plan
- [x] `make lint`
- [x] `python -m unittest application.tests.pgvector_utils_test application.tests.prompt_client_pgvector_similarity_test application.tests.librarian.candidate_retriever_test`
- [x] Vec-only coverage in `db_test` (+ sync/rewrite script unit tests)
- [x] Local Postgres: downgrade parents → upgrade `c7d8e9f0a1b2` → 4029/4029 `embedding_vec`, CSV gone
- [ ] CI green
- [ ] Staging/prod: backup → `flask db upgrade` (CSV→vector then DROP CSV) → smoke chat + optional `CRE_LIBRARIAN_RETRIEVER_BACKEND=pgvector`

## Out of scope
- Prod `alembic_version` surgery / rewriting historical migrations
- Removing numpy/scipy/sklearn from prod requirements (#978 follow-on)
- HNSW / halfvec ANN index (Essential-1 capacity; track separately)
- Dropping SQLite as a supported app DB entirely — maintainer follow-up #982 (post-GSoC / as part of that merge)

PR #837, opened 2026-03-29
Fix variable shadowing crash in get_by_tags and code quality improvements

BEAST04289:
## Summary
Fixes a crash bug and several code quality issues found during codebase review for GSoC 2026 Module B proposal.

Closes #836 
### Bug Fix: Variable shadowing crash in get_by_tags() (application/database/db.py)
The loop variable node was overwritten by the return value of self.get_nodes(). When get_nodes() returned [] (empty list), the error handler tried to access node.name on the empty list, causing AttributeError. Renamed to db_node and resolved to preserve the original DB node reference for error logging.

### Code Quality Fixes
- Unreachable code in add_node() (db.py:1585): Removed dead return None after raise ValueError
- - Duplicate import (web_main.py:15,20): Removed redundant from application.utils import oscal_utils, redis
- - Variable typo (gap_analysis.py:44-46): Renamed pentalty to penalty
- - Variable typo (cre_main.py:280): Renamed pending_stadards to pending_standards (4 occurrences)
## Testing
- All changes are localized and do not alter behavior
- - The variable shadowing fix preserves the original node reference for correct error logging
-

PR #980, opened 2026-07-13
feat(db): persist users + resource selection — Part of #586 

skypank-coder:
## What & why
First increment of #586 ("Add support for users"): login-first user
persistence. OIDC login existed but nothing was persisted — no User table,
nowhere to hang per-user data. This adds that foundation so per-user resource
filtering (PR2–PR4) can be built on it. Aligns with auth RFC #876 (TODO 1/2).

## Changes
- **`users`** table anchored on the immutable OIDC `sub` (`google_sub`); email
  and display_name refreshed on each login.
- **`user_resource_selection`** table: one row per selected standard,
  `UNIQUE(user_id, standard_name)`, `FK → users.id ON DELETE CASCADE`.
- **Login callback** upserts the User and sets `session['user_id']` — gated on
  `CRE_ENABLE_LOGIN`, wrapped so it never blocks login. Existing session keys
  and the anonymous 401 are unchanged (intentionally narrower than RFC #876's
  `login_required` rewrite).
- **`Node_collection`** methods: `upsert_user`, `get_user_by_sub`,
  `get_user_resource_selection`, `set_user_resource_selection`. These back the
  `/rest/v1/user` resource endpoints coming in PR2.
- **Migrations:** a dedicated merge revision collapses the two pre-existing
  open heads (`ab12cd34ef56`, `9b1c2d3e4f50`), then a revision creates the two
  tables — keeping the head single-parent so `flask db downgrade` is
  unambiguous.

## Testing
- 12 new tests: idempotent upsert (no dup row), selection round-trip / replace /
  dedup / per-user isolation, unique constraint, cascade delete, callback upsert,
  and the flag-off no-op path.
- Verified on **real Postgres** (prod parity), not just SQLite: single head,
  `ON DELETE CASCADE` enforced at the DB level (rows 1 → 0), both unique
  constraints created cleanly, migration `upgrade → downgrade → upgrade`
  deterministic, alembic revision guardrail green. 12/12 pass on Postgres.
- `black` clean; no new mypy errors vs. baseline.

## Scope boundaries
No new endpoints, no server-side filtering, no frontend, no `login_required`
rewrite — those are the follow-up PRs in #586.

## Unrelated pre-existing issue (out of scope, flagging for visibility)
`flask db upgrade` from a **blank** DB fails on Postgres before reaching my
revisions: migrations `3c65127871a6` / `7bf4eac76958` reuse the constraint name
`uq_pair` (authored via SQLite `batch_alter_table`), which collides on a
from-base replay. Postgres rolls the whole thing back cleanly (transactional
DDL). This predates this branch and doesn't affect the prod path (apply-onto-
existing-schema), so it's out of scope here — but worth a separate fix, since
anyone bootstrapping a fresh Postgres DB will hit it.


migrations/versions/f0e1d2c3b4a5_merge_heads_before_users.py
```diff
@@ -0,0 +1,27 @@
...
+Create Date: 2026-07-08
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "f0e1d2c3b4a5"
+down_revision = ("ab12cd34ef56", "9b1c2d3e4f50")
```
northdpole: **Blocking:** `main` already merges `ab12cd34ef56` + `9b1c2d3e4f50` via `c7d8e9f0a1b2` (pgvector, #979). This second merge of the same parents leaves two Alembic heads after merge:

- `c7d8e9f0a1b2`
- `a1b2c3d4e5f6`

Please drop this file entirely and point the users migration at `c7d8e9f0a1b2` instead.

Note: the deploy alembic guardrail only checks DB revision ∈ known revisions — it will not catch this dual-head case. Confirm with `flask db heads` (single head) after the fix.

migrations/versions/a1b2c3d4e5f6_add_users_and_resource_selection.py
```diff
@@ -0,0 +1,50 @@
...
+Create Date: 2026-07-08
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "a1b2c3d4e5f6"
+down_revision = "f0e1d2c3b4a5"
```
northdpole: **Blocking (paired with the obsolete merge revision):** after dropping `f0e1d2c3b4a5`, set `down_revision = "c7d8e9f0a1b2"` and update the docstring `Revises:` line to match.

application/web/web_main.py
```diff
@@ -1219,6 +1219,27 @@ def callback():
...
+            )
+        else:
+            try:
+                user = db.Node_collection().upsert_user(
+                    google_sub=google_sub,
+                    email=id_info.get("email") or "",
+                    display_name=id_info.get("name"),
+                )
+                session["user_id"] = user.id
+            except Exception as e:
```
northdpole: Nit (non-blocking): catching bare `Exception` is fine for "never block login", but narrowing to `SQLAlchemyError` (or similar) would still keep DB failures soft while letting unexpected bugs surface. Up to you for this PR.

northdpole (changes requested):
Thanks — this is a clean foundation for #586: idempotent upsert with race recovery, cascade + uniqueness covered, callback gated so persistence can't break login, and solid Postgres-minded tests.

**Blocking:** migration lineage conflicts with current `main` (see inline). After that fix I'm happy to approve.

**Also (not in the diff, so calling out here):** `NO_LOGIN=1` in `/rest/v1/login` still only sets `google_id` / `name`, not `session["user_id"]`, and never upserts a User. Local/dev will look logged in while PR2's `/rest/v1/user/resources` has nothing to hang selection on. Please either fix here or explicitly handle in #981.

Once this lands, please rebase #981 onto updated `main` (or stack it on this branch) so we don't re-review the DB layer.

skypank-coder:
@northdpole Thanks for the thorough review — all three points addressed in fd5bd59.

Migration lineage (blocking). You were right, and I reproduced it before fixing: flask db heads reported both a1b2c3d4e5f6 and c7d8e9f0a1b2. Since c7d8e9f0a1b2 already carries down_revision = ("9b1c2d3e4f50", "ab12cd34ef56"), my revision was a duplicate merge of the same two parents.

Deleted f0e1d2c3b4a5_merge_heads_before_users.py entirely
Pointed the users migration at down_revision = "c7d8e9f0a1b2" and updated the Revises: docstring
Confirmed on Postgres after the fix: flask db heads → single head a1b2c3d4e5f6; upgrade → downgrade → upgrade clean; alembic guardrail OK. Also noted your point that the guardrail only checks revision membership and wouldn't catch the dual-head case — I used the head count as the real gate.

NO_LOGIN dev path. Good catch — fixed here rather than deferring to #981, since it belongs with the persistence work. /rest/v1/login now upserts the dev User and sets session["user_id"] when CRE_ENABLE_LOGIN is on, so local/dev sessions have a real row for user-scoped endpoints. Two tests cover it: persists with the flag on, no-op with it off.

Nit — bare except. Adopted. Narrowed to SQLAlchemyError in both the callback and the new dev path, so DB failures stay soft while unexpected non-DB bugs surface.

On #981. It's stacked directly on this branch (verified this branch's tip is an ancestor of it). Once this merges I'll rebase #981 onto updated main so the DB layer drops out of its diff and you only review the resource-selection endpoint.

Ready for another look.

northdpole (approved):
Re-reviewed after fd5bd59 — all three prior points addressed:

- obsolete merge revision dropped; users migration chains off `c7d8e9f0a1b2` (single head)
- `NO_LOGIN` path upserts User + sets `session["user_id"]` when login is enabled (with flag-off coverage)
- persistence failures narrowed to `SQLAlchemyError`

CI green. LGTM. Please merge/rebase `main` (just #837) before merge if strict up-to-date checks require it, then rebase #981 onto updated main.

skypank-coder:
@northdpole Branch is updated with #837 and CI's green — ready to merge whenever you are. I'll rebase #981 onto main right after it lands.

Issue #966, opened 2026-07-08
Auth-wrap POST /rest/v1/cre_csv_import (deprecate anonymous MyOpenCRE import)


northdpole:
## Context

RFC #876 adds authenticated MyOpenCRE upload at `POST /rest/v2/myopencre/upload`. Today, `POST /rest/v1/cre_csv_import` is **public** (no auth) and writes directly into the shared graph via `myopencre_parser` — intended for local/self-hosted use.

Maintainer decision: **`cre_csv_import` must be auth-wrapped** for online deployment. Anonymous CSV import into the shared CRE graph is not acceptable on opencre.org.

## Requirements

- [ ] Add `@login_required` (or equivalent post-RFC TODO 1 session check) to `POST /rest/v1/cre_csv_import`
- [ ] Return 401/302 for anonymous requests (consistent with auth route migration — see auth `/rest/v1/auth/*` issue)
- [ ] Document that v1 import remains for **authenticated admin/contributor** use; per-user mapping flow is `/rest/v2/myopencre/upload` (RFC TODO 4)
- [ ] Frontend MyOpenCRE page: if using v1 import path today, switch to v2 upload when available or require login for v1
- [ ] Consider gating behind `CRE_ALLOW_IMPORT` and/or `capabilities.myopencre` in addition to login
- [ ] Update tests in `web_main_test.py` / MyOpenCRE frontend tests

## Migration notes

- Local dev with `NO_LOGIN=1` should continue to work for contributor workflows
- `GET /rest/v1/cre_csv` template download may stay public (read-only) — confirm during implementation

## Related

- `docs/rfc/user-auth-myopencre.md`
- #586
- `application/web/web_main.py` `import_from_cre_csv`

skypank-coder:
@northdpole I'd like to take this too — it pairs naturally with #963, since the auth-wrapping here should use the same TODO 1 session check I'll be adding in that migration. I'll sequence it right after (or alongside) #963 so the 401/302 behavior is consistent between them.
One thing I'll confirm during implementation per your note: keeping GET /rest/v1/cre_csv (template download) public since it's read-only, and preserving NO_LOGIN=1 for local dev. I'll gate the import behind login plus the existing CRE_ALLOW_IMPORT / capabilities.myopencre flags, and cover both anonymous-rejected and authenticated-allowed cases in the tests.

northdpole:
Thanks for volunteering on this — assigning to you.

[...]

Issue #963, opened 2026-07-08
Auth: migrate OAuth routes from /rest/v1/login to /rest/v1/auth/*


northdpole:
## Context

RFC merged in #876 (`docs/rfc/user-auth-myopencre.md`) proposes new auth routes at `/auth/login`, `/auth/callback`, `/auth/logout`. Existing Google OAuth already lives at:

- `GET /rest/v1/login`
- `GET /rest/v1/callback`
- `GET /rest/v1/logout`
- `GET /rest/v1/user`

Per maintainer review of #876, consolidate under the versioned API namespace:

**Target routes:**
- `GET /rest/v1/auth/login`
- `GET /rest/v1/auth/callback`
- `GET /rest/v1/auth/logout`
- `GET /rest/v1/auth/user` (or keep `/rest/v1/user` — decide during implementation)

## Requirements

- [ ] Add `/rest/v1/auth/*` routes implementing the RFC TODO 1 session contract (`session['user_id']`, redirect on missing session)
- [ ] Keep `/rest/v1/login`, `/rest/v1/callback`, `/rest/v1/logout` as **temporary aliases** with deprecation headers or redirect to `/rest/v1/auth/*`
- [ ] Update frontend (`useUser`, Header) and any docs referencing old paths
- [ ] Update `login_required` to redirect to `/rest/v1/auth/login?next=<url>` (not 401 for browser flows)
- [ ] Consider separate behavior for API clients (`Accept: application/json` → 401 vs browser → 302)

## Related

- RFC: `docs/rfc/user-auth-myopencre.md` TODO 1
- #586 (MyOpenCRE online)
- #943 (frontend auth UX)
- #949 (capabilities API)

## Acceptance criteria

- New routes work end-to-end with Google OAuth
- Old routes redirect or alias without breaking existing bookmarks
- Tests cover both old aliases and new canonical paths during migration window

skypank-coder:
@northdpole I'd like to take this — it's TODO 1 of the RFC and builds directly on #943 and #949, so it's a natural next step for me.
My read on the approach:

Add the /rest/v1/auth/* routes implementing the TODO 1 session contract (session['user_id'], redirect on missing session), with the old /rest/v1/login|callback|logout kept as aliases during the migration window so existing bookmarks don't break.
Migrate login_required to redirect browser flows to /rest/v1/auth/login?next=<url>, while keeping a 401 for API clients (Accept: application/json) — matching the content-negotiation point in the requirements.
[...]

PR #981, opened 2026-07-13
feat(api): per-user resource selection endpoint — Part of #586

skypank-coder:
## What & why
Second increment of #586. Builds on PR1's user persistence to expose the
per-user resource selection over the API, so the frontend (PR4) and server-side
filtering (PR3) have a read/write surface. No new tables or migrations — pure
HTTP + validation over PR1's `Node_collection` methods.

## Endpoints
`/rest/v1/user/resources` (canonical `/rest/v1` surface):
- **`GET`** → `{"selected": [...]}` — the current user's saved standard names, sorted.
- **`PUT`** `{"selected": ["ASVS","CWE"]}` → dedup + replace, persists via PR1's
  `set_user_resource_selection`, returns the stored list.

## Gating
Requires `login_required` + `is_login_enabled()` + `is_myopencre_enabled()`
(per the #966 decision):
- Login flag off → safe default `{"selected": []}`, 200, no auth, no writes.
- MyOpenCRE capability off → same safe default; the gate short-circuits before
  any DB read/write.
- Flag on + anonymous → 401.
- Invalid PUT body (missing `selected`, not a list, empty/non-string entries) → 400.

Implemented via a generalized `feature_enabled_or_default` helper composed over
`login_required` (a generalization of PR1's flag-gating; PR1's behaviour is
unchanged).

## OpenAPI
Endpoint added to the spec properly — two `PathSpec` entries +
`@openapi_documented` decorators, `docs/api/openapi.yaml` regenerated (not
hand-edited). The OpenAPI guardrail passes all four checks (documented views,
freshness, validity, route coverage).

## Testing
12 new tests, verified on **real Postgres** (prod parity): flag-off default,
myopencre-off default + no-write, anonymous 401, dedup, replace against the live
`UNIQUE(user_id, standard_name)`, and body-validation 400s. The myopencre-off
tests are mutation-proven — removing the capability from the gate makes them fail
(GET leaks the saved selection, PUT persists), restoring it makes them pass — so
they genuinely detect a bypassed gate rather than passing incidentally.

`black` clean; no new mypy errors vs. baseline.

## Scope boundaries
No server-side filtering (PR3), no frontend (PR4), no new DB surface. Reuses
PR1's methods and `login_required`.

## Files (4)
- `application/web/web_main.py` — routes + gate + `feature_enabled_or_default`.
- `application/web/openapi_registry.py` — two `PathSpec` entries.
- `docs/api/openapi.yaml` — regenerated.
- `application/tests/user_resources_api_test.py` — new, 12 tests.

### DIFF
--- a/application/tests/user_resources_api_test.py
+++ b/application/tests/user_resources_api_test.py
@@ -17,6 +17,7 @@ from application.database import db
 class TestUserResourcesApi(unittest.TestCase):
     def setUp(self) -> None:
         # SQL-only surface; skip the Neo4j graph load and allow http in tests.
+        self._prev_no_load_graph = os.environ.get("NO_LOAD_GRAPH_DB")
         os.environ["NO_LOAD_GRAPH_DB"] = "1"
         self.app = create_app(mode="test")
         self.app.secret_key = "test-secret"
@@ -29,7 +30,12 @@ class TestUserResourcesApi(unittest.TestCase):
         sqla.session.remove()
         sqla.drop_all()
         self.app_context.pop()
-        os.environ.pop("NO_LOAD_GRAPH_DB", None)
+        # Restore the prior value rather than unconditionally deleting it, so a
+        # value set by the test runner survives for later tests.
+        if self._prev_no_load_graph is None:
+            os.environ.pop("NO_LOAD_GRAPH_DB", None)
+        else:
+            os.environ["NO_LOAD_GRAPH_DB"] = self._prev_no_load_graph
 
     def _login(self, client: Any, google_sub: str = "sub-1", name: str = "U") -> None:
         with client.session_transaction() as sess:
--- a/application/tests/user_resources_api_test.py
+++ b/application/tests/user_resources_api_test.py
@@ -1,0 +1,338 @@
+"""Tests for the per-user resource-selection API (issue #586, PR2).
+
+GET/PUT /rest/v1/user/resources, gated by login_required + is_login_enabled.
+Flag-off returns a safe default; authenticated users read/write their selection.
+"""
+
+import json
+import os
+import unittest
+from typing import Any
+from unittest.mock import patch
+
+from application import create_app, sqla
+from application.database import db
+
+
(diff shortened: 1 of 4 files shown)


northdpole (changes requested):
Thanks for the clean API surface — `feature_enabled_or_default` + login/MyOpenCRE gate, validation/trim, OpenAPI coverage, and the mutation-proven flag-off tests look solid.

**Blocking before approve:** rebase onto current `main` (now includes #980). This branch still carries the pre-#980 DB layer:

1. **Obsolete Alembic merge** — still has `f0e1d2c3b4a5` and `a1b2c3d4e5f6` revising it. On `main`, users tables land via `a1b2c3d4e5f6` → `c7d8e9f0a1b2` (no `f0e1…`). After rebase this PR should be **API-only** (routes, OpenAPI, `user_resources_api_test.py`) with no migration files / no re-introducing the DB models.
2. **Stale login paths vs #980** — this branch’s `/rest/v1/login` `NO_LOGIN` path and broad `except Exception` on callback persistence are behind what landed in #980. Rebase will pick those up; don’t re-apply the old versions.

**Non-blocking (nice after rebase):**
- Global `@app.after_request` sets `cache_control.max_age = 300`; consider `Cache-Control: no-store` on the per-user GET/PUT responses.
- Prefer `session['user_id']` when present in `_resolve_current_user` before upsert-by-sub (avoids an extra lookup on the happy path after #980).

Once rebased green, ping for re-review — then #963 (auth route migration) can open cleanly on top of the persisted-user foundation.

northdpole (approved):
Re-reviewed after the rebase onto main/#980.

Blocking items from the earlier review are addressed: API-only diff (no migrations / no reintroduced DB layer), current login paths preserved, and the non-blocking `no-store` + `session['user_id']` preferences landed. CI is green.

Optional follow-ups only (not blocking): assert GET ordering without re-sorting in the persist test, and seed a selection before invalid PUT 400s to prove no partial replace.

Approving and merging with rebase.

PR #985, opened 2026-07-18
GSoC Module A : Week 3: feat(harvester): implement incremental change detection (stacked on #983)

ParthAggarwal16:
## Summary

(this PR is stacked on top of #983 )
This PR introduces change detection and checkpoint persistence for the harvester pipeline.

It adds Git-based change detection utilities for discovering modified files and commits since a checkpoint, along with a lightweight checkpoint store for persisting repository processing state across pipeline runs.

**Note:** This PR is part of the Week 3 implementation and is intended to be reviewed after **Week 2 (Repository Client)**, even though it targets `main`.

## Added

- `ChangeDetector` for Git-based incremental change detection
- Modified file discovery since a commit
- Commit history retrieval since a checkpoint
- JSON-backed `CheckpointStore`
- Repository checkpoint persistence
- Repository checkpoint loading
- `RepositoryCheckpoint` model
- `RepositoryChangeSet` model
- Unit tests for change detection
- Unit tests for checkpoint persistence

### Change Detection

- Modified file discovery
- Duplicate file filtering
- Commit history retrieval
- Commit ordering validation

### Checkpoint Store

- Save checkpoint
- Load existing checkpoint
- Missing checkpoint file handling
- Missing repository checkpoint handling

## Test Plan

Executed:

```bash
python3 -m unittest application.tests.harvester_test.change_detector_test
python3 -m unittest application.tests.harvester_test.checkpoint_store_test
python3 -m unittest discover -s application/tests/harvester_test -p "*_test.py"
make test
```

All tests passing.

## Notes for Reviewers

- Checkpoints are stored as simple JSON to keep the implementation lightweight and easy to inspect during development.
- `ChangeDetector` is intentionally separated from the repository client so repository operations and incremental change detection remain independent concerns.
- Duplicate modified files are normalized before being returned to downstream pipeline stages.
- This PR establishes the incremental harvesting foundation used by the following pipeline stages.

1: High-Level Architecture (HLA)

image

2. Class Diagram
image

3. Data Flow Diagram — get_modified_files_since

image

4. Data Flow Diagram — get_commits_since
image

5. Sequence Diagram — Checkpoint Save/Load
image

6. End-to-End Incremental Harvest Flow (conceptual, based on the two components together)
image

PR #986, opened 2026-07-18
GSoC Module A : week 4 : feat(harvester): implement repository file filtering (stacked on top of #985)

ParthAggarwal16:
## Summary

(this PR is stacked on top of #985 )
This PR adds the file filtering layer for the harvester pipeline, ensuring that only relevant documentation files continue through the harvesting process.

> **Note:** This PR depends on the previous stacked PRs (Week 2 and Week 3) and should be reviewed after them.

## Added

- File filtering based on supported documentation extensions
- Regex-based exclusion rules for common non-documentation paths
- Configurable allowlist of supported file extensions
- Configurable exclude pattern support
- Filtering metrics collector
- Filtering benchmark utilities
- Filtering benchmark result model
- Unit tests covering filtering behavior and metrics

### Allowed extensions

- `.md`
- `.mdx`
- `.rst`
- `.txt`
- `.adoc`

### Default excluded paths

- `.github/`
- `.git/`
- `node_modules/`
- `dist/`
- `build/`
- `coverage/`
- `vendor/`
- `package-lock.json`
- `yarn.lock`
- `pnpm-lock.yaml`

### File Filtering

- Extension allowlist filtering
- Regex-based path exclusion
- Combined extension + regex filtering

### Metrics

- Retained file counting
- Filtered file counting
- Total file counting

### Benchmark

- Retention rate calculation
- Filtering rate calculation
- Benchmark summary generation

## Test Plan

Executed:

```bash
python3 -m unittest discover \
    -s application/tests/harvester_test \
    -p "*_test.py"

make test
```

All tests passing.

## Notes for Reviewers

- Filtering is intentionally separated from change detection to keep responsibilities isolated.
- Default filtering rules are configurable through the `FileFilter` constructor.
- Benchmark and metrics utilities are lightweight helpers intended for future pipeline instrumentation.

1. Class Diagram
image

2. Sequence Diagram — filter_files() call flow
image

3. Sequence Diagram — FilteringBenchmark.run() call flow
image

4. Data Flow Diagram — Benchmark metrics aggregation
image

5. High-Level Architecture (Component Diagram) — simplified
image

6. Data Flow Diagram — end-to-end filtering pipeline — simplified
image

### DIFF
--- a/application/tests/harvester_test/file_filter_test.py
+++ b/application/tests/harvester_test/file_filter_test.py
@@ -1,0 +1,62 @@
+import unittest
+
+from application.utils.harvester.file_filter import (
+    FileFilter,
+)
+
+
+class FileFilterTests(unittest.TestCase):
+    def test_extension_filtering(self):
+        file_filter = FileFilter()
+
+        result = file_filter.filter_files(
+            [
+                "README.md",
+                "image.png",
+                "script.js",
+            ]
+        )
+
+        self.assertEqual(
+            result,
+            ["README.md"],
+        )
+
+    def test_regex_filtering(self):
+        file_filter = FileFilter()
+
+        result = file_filter.filter_files(
+            [
+                ".github/workflows/test.yml",
+                "docs/setup.md",
+            ]
+        )
+
+        self.assertEqual(
+            result,
+            ["docs/setup.md"],
+        )
+
+    def test_combined_filtering(self):
+        file_filter = FileFilter()
+
+        result = file_filter.filter_files(
+            [
+                "README.md",
+                ".github/workflows/test.yml",
+                "node_modules/react/index.js",
+                "docs/setup.md",
+            ]
+        )
+
+        self.assertEqual(
+            result,
+            [
+                "README.md",
+                "docs/setup.md",
+            ],
+        )
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/application/tests/harvester_test/file_filter_test.py
+++ b/application/tests/harvester_test/file_filter_test.py
@@ -57,6 +57,31 @@ class FileFilterTests(unittest.TestCase):
             ],
         )
 
+    def test_empty_overrides_are_respected(self):
+        file_filter = FileFilter(
+            exclude_patterns=[],
+            allowed_extensions=set(),
+        )
+
+        result = file_filter.filter_files(
+            [
+                "README.md",
+                "image.png",
+            ]
+        )
+
+        self.assertEqual(result, [])
+
+    def test_default_instances_are_isolated(self):
+        first = FileFilter()
+        second = FileFilter()
+
+        first.exclude_patterns.append("custom")
+        first.allowed_extensions.add(".pdf")
+
+        self.assertNotIn("custom", second.exclude_patterns)
+        self.assertNotIn(".pdf", second.allowed_extensions)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/application/tests/harvester_test/file_filter_test.py
+++ b/application/tests/harvester_test/file_filter_test.py
@@ -112,6 +112,10 @@ class FileFilterTests(unittest.TestCase):
             second.exclude_patterns,
         )
 
+    def test_empty_extension_raises(self):
+        with self.assertRaises(ValueError):
+            FileFilter(allowed_extensions={""})
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/application/tests/harvester_test/file_filter_test.py
+++ b/application/tests/harvester_test/file_filter_test.py
@@ -22,7 +22,7 @@ class FileFilterTests(unittest.TestCase):
             ["README.md"],
         )
 
-    def test_regex_filtering(self):
+    def test_path_exclusion(self):
         file_filter = FileFilter()
 
         result = file_filter.filter_files(
@@ -43,8 +43,8 @@ class FileFilterTests(unittest.TestCase):
         result = file_filter.filter_files(
             [
                 "README.md",
-                ".github/workflows/test.yml",
-                "node_modules/react/index.js",
+                ".github/workflows/README.md",
+                "node_modules/react/README.md",
                 "docs/setup.md",
             ]
         )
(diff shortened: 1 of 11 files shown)


application/utils/harvester/checkpoint_store.py
```diff
@@ -0,0 +1,64 @@
...
+        temp_file = self.checkpoint_file.with_suffix(".tmp")
+        temp_file.write_text(
+            json.dumps(
+                data,
+                indent=2,
+            ),
+            encoding="utf-8",
+        )
+
+        os.replace(temp_file, self.checkpoint_file)
```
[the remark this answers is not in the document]
ParthAggarwal16: wrong, harvester currently isn’t multi-process or multi-threaded.
Current implementation assumes a single harvester process, so a fixed temporary file is sufficient for now. We can revisit unique temporary files if concurrent checkpoint writers are introduced.

application/utils/harvester/file_filter.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(
+        self,
+        exclude_patterns: list[str] | None = None,
+        allowed_extensions: set[str] | None = None,
+    ):
+        self.exclude_patterns = exclude_patterns or DEFAULT_EXCLUDE_PATTERNS
+        self.allowed_extensions = allowed_extensions or DEFAULT_ALLOWED_EXTENSIONS
+
+    def is_excluded_by_pattern(self, file_path: str) -> bool:
+        return any(re.search(pattern, file_path) for pattern in self.exclude_patterns)
```
northdpole: **Blocker: this API receives glob patterns but interprets them as regular expressions.** `PathRules.exclude`, `repos.yaml`, and `exclude_patterns.txt` define values such as `**/archive/**`; feeding that value to `re.search()` can raise `re.error`, so the expected config-to-filter path is unusable. Apply the fix by making globs the single public contract: normalize each repository-relative path to POSIX separators, validate exclusion globs during configuration loading/construction, and match them with one glob-aware helper rather than `re.search()`. Wire `repository.paths.exclude` through that helper and add a regression test using the actual `**/archive/**` configuration. Do not maintain a second regex syntax for the same setting.

application/utils/harvester/file_filter.py
```diff
@@ -0,0 +1,54 @@
...
+
+DEFAULT_ALLOWED_EXTENSIONS = {
+    ".md",
+    ".mdx",
+    ".rst",
+    ".txt",
+    ".adoc",
+}
+
+DEFAULT_EXCLUDE_PATTERNS = [
```
northdpole: These defaults duplicate—but do not match—the existing exclusion configuration. Root anchors miss nested directories, and the list omits configured paths such as `archive`, `.cursor`, and `.claude`. Apply the fix by defining one canonical immutable set of default **globs**, then combine it with each repository's configured exclusions through the same matching path. Ensure directory exclusions apply at repository root and any nesting depth. Add allowed-extension cases for `.github/README.md`, `packages/site/node_modules/README.md`, `docs/archive/old.md`, and `.cursor/rules/project.md`. The partial nested-pattern adjustment in #987 does not resolve the duplicated contract or all omitted paths.

application/utils/harvester/file_filter.py
```diff
@@ -0,0 +1,54 @@
...
+]
+
+
+class FileFilter:
+    def __init__(
+        self,
+        exclude_patterns: list[str] | None = None,
+        allowed_extensions: set[str] | None = None,
+    ):
+        self.exclude_patterns = exclude_patterns or DEFAULT_EXCLUDE_PATTERNS
```
northdpole: Using `or` discards intentional empty overrides, while assigning the module-level list/set directly shares mutable state between instances. Apply the fix by distinguishing absence explicitly: use a copy of defaults only when the argument is `None`; otherwise copy the supplied collection even when it is empty. Do the same for `allowed_extensions` (for example, `list(defaults) if exclude_patterns is None else list(exclude_patterns)` and the equivalent `set(...)`). Add tests proving empty overrides remain empty and mutations in one `FileFilter` instance cannot affect another.

application/tests/harvester_test/file_filter_test.py
```diff
@@ -0,0 +1,62 @@
...
+            result,
+            ["README.md"],
+        )
+
+    def test_regex_filtering(self):
+        file_filter = FileFilter()
+
+        result = file_filter.filter_files(
+            [
+                ".github/workflows/test.yml",
```
northdpole, line 30: This does not test exclusion matching because `.yml` is already rejected by the extension allowlist; the combined test has the same problem with `.js`. Replace excluded examples with allowed documentation extensions so removing exclusion matching would make the test fail, e.g. `.github/workflows/README.md` and `node_modules/react/README.md`. Add focused cases for nested paths, the real `**/archive/**` custom glob, invalid glob handling, explicit empty exclusions/extensions, and default-instance isolation. Keep extension and path-filter behavior in separate tests so each assertion has one reason to pass.

northdpole:
**Status: not mergeable** (no push from us).

Open blockers still apply:
- Glob patterns treated as regexes
- Default excludes drift from \`exclude_patterns.txt\` / YAML
- Mutable shared defaults via \`or\`
- Exclusion tests don’t actually hit the exclude path

Also: this branch is on the **stale** pre-#985 week-3 tip. Rebase onto the fixed #985 tip after that lands, then address filter blockers.

northdpole:
Still blocked: no new commits since the filter blockers, and branch still conflicts / sits on the old week-3 tip. After #985 is fixed and merged, please rebase \`week_4-clean\` onto that tip and address the open \`file_filter\` threads.

application/tests/harvester_test/file_filter_test.py
ParthAggarwal16: i looked into invalid glob handling, i tried using malformed patterns such as [ expecting pathspec to reject them, but pathspec.PathSpec.from_lines() accepts them without raising an exception, so I couldn’t find a practical invalid glob case to test against, is there a specific pattern you had in mind?
image

ParthAggarwal16:
hey, i think i have addressed everything, good to go from my side, let me know if you need any further improvments
@northdpole

ParthAggarwal16:
rebased and good to go ᕦ(⩾﹏⩽)ᕥ

northdpole:
**Please rebase onto latest `main`** (~2 commits behind after recent merges). #985 is already merged, so this should sit cleanly on `main` as the Module A week-4 base.

Once rebased + green, ping for review — next in the Module A queue before #987.

northdpole (commented):
## Preliminary review — Module A week 4 (#986)

Looks like the right next Module A merge after #985. CI green; please **rebase onto `main`** (~2 behind) so the diff is week-4-only against current main, then I’ll do a full approve pass on `file_filter` / metrics / tests.

No code objections from the file list/PR description so far — rebase first so we don’t review a stale stack tip.

northdpole (approved):
## Review — Module A week 4 file filtering (#986)

CI green again after re-running the Test job that died in yesterday’s Actions outage (`Failed to resolve action download info`). Ignoring perpetual e2e noise.

### Looks good
- Clear allowlist (docs extensions) + gitignore-style excludes via `pathspec`
- Defaults loaded from `exclude_patterns.txt` with override hooks for tests
- Small metrics collector + hermetic unit tests (extension, path exclusion, nested globs, empty overrides)
- Sits cleanly on merged #985 as week-4 base

### Non-blocking
- `is_allowed_extension` uses `endswith` on the raw path; `Path(file_path).suffix.lower()` would be slightly safer for case variants
- Confirm `pathspec` is already a declared dependency on `main` (import is used; CI passed so it’s available in the install set)

Approving as the Module A week-4 merge candidate. #987 can rebase on this once landed.

northdpole:
Yesterday’s failed Test run on this branch was the **GitHub Actions outage** (action download `Service Unavailable`). Re-ran CI — current head is green.

PR #988, opened 2026-07-20
fix(migrations): stop 7bf4eac76958 renaming link constraints to a sha…

manshusainishab:
## Problem

A from-scratch `flask db upgrade` on PostgreSQL fails:

    psycopg2.errors.DuplicateTable: relation "uq_pair" already exists
    [SQL: ALTER TABLE cre_node_links ADD CONSTRAINT uq_pair UNIQUE (cre, node)]

Migration `7bf4eac76958` renames the unique constraints on **both** `cre_links`
(`uq_cre_link_pair`) and `cre_node_links` (`uq_cre_node_link_pair`) to the same
name `uq_pair`. On Postgres a UNIQUE constraint creates a schema-global index,
so the second `uq_pair` collides with the first — breaking any from-empty
migration on Postgres. (It slips by on SQLite, where `batch_alter_table`
rebuilds the table.)

## Fix

Remove those two renames from `7bf4eac76958` (upgrade + downgrade). The preceding
migration `0d267ae11945` already creates the table-specific constraints, and the
SQLAlchemy models keep those names (`uq_cre_link_pair` / `uq_cre_node_link_pair`),
so we preserve them instead of collapsing both to `uq_pair`. The migration's
actual purpose (adding `node.section_id` and updating `uq_node`) is unchanged.

## Validation

Full `flask db upgrade` from an empty PostgreSQL database now completes the entire
chain to head `c7d8e9f0a1b2`.

Note: completing the run also requires `CRE_EMBED_EXPECTED_DIM` to be set for the
pgvector migration (`c7d8e9f0a1b2`) on an empty DB — a separate, pre-existing
requirement of that migration, unrelated to this fix.


manshusainishab:
@northdpole here is the PR with the fix as discussed I need a review and merge so I can take a fresh pull to continue with my next PR.

northdpole (approved):
Thanks for tracking this down and fixing it cleanly.

Approved: removing the dual rename to `uq_pair` is the right fix for empty Postgres `flask db upgrade` (schema-global unique indexes), upgrade/downgrade stay consistent, and the real `section_id` / `uq_node` work is untouched. Models already use `uq_cre_link_pair` / `uq_cre_node_link_pair`.

Merging with rebase — appreciate the careful migration surgery.

Issue #599, opened 2025-05-24
Unable to Access OpenCRE Google Sheet – Permission Denied


althunibat:
Hi OpenCRE team,

We’ve been trying to access the spreadsheet at:

https://docs.google.com/spreadsheets/d/1eZOEYgts7d_-Dr-1oAbogPfzBLh6511b58pX3b59kvg

We followed the instructions and attempted access using valid OAuth desktop app credentials, but we continue to receive a 403: The caller does not have permission error.

Could you please either:
	•	Share read-only access to the sheet, or
	•	Provide a static export (preferably in JSON)?

Thank you!

northdpole:
To get the same info in your local database try the upstream sync cli option.
Thank you for using opencre!

northdpole:
You need to run migrations first 
Make migrate-upgrade

paoga87:
@northdpole I had run `make migrate-upgrade` first... that's why it was odd getting the error above.

northdpole:
Hi @althunibat — thanks for raising this, and thanks @paoga87 for trying the upstream path.

You **do not need access to the Google Sheet**. That spreadsheet is an internal maintainer workflow for bulk standards imports. For contributors and local development, the supported path is:

1. **Pull the OpenCRE graph from upstream (public API → local DB)**
2. **Import your own mappings via MyOpenCRE** (CSV upload against your local instance)

---

## 1. Get a local database from upstream

Setup guide: https://github.com/OWASP/OpenCRE/blob/main/docs/developmentSetup.md

Typical first-time flow from repo root:

```bash
make install
make migrate-upgrade          # create schema first (required)
make upstream-sync            # downloads the CRE graph from https://opencre.org via REST
```

That uses `python cre.py --upstream_sync`, which pulls root CREs and linked data from the public API (`/rest/v1/root_cres`, `/rest/v1/id/{creid}`) into `standards_cache.sqlite` by default.

If you see `no such table: cre` during upstream sync, migrations were not applied yet — run `make migrate-upgrade` first (as noted in the thread).

Optional: point at a different upstream with `CRE_UPSTREAM_API_URL` (defaults to `https://opencre.org/rest/v1`).

Then start services and the app locally:

```bash
make start-containers   # Redis + Neo4j
make dev-flask          # backend (default http://localhost:5000)
# in application/frontend: yarn start  → http://localhost:9001
```

---

## 2. Import **your own** standard mappings (MyOpenCRE)

[...]

PR #997, opened 2026-07-28
docs: clarify local contributor setup for upstream sync and MyOpenCRE

Bornunique911:
## Summary
This PR updates the contributor documentation for issue #599  to clarify the supported local development workflow for OpenCRE.

In particular, it makes it explicit that contributors do not need access to the internal OpenCRE Google Sheet for normal local development, and that the supported bootstrap path is to create the local schema, sync from upstream, and optionally use local MyOpenCRE / CSV import when needed.

## Problem
The current setup documentation leaves a gap around how contributors should work locally when they do not have access to the internal OpenCRE spreadsheet workflow.

Because of that, it is easy to assume that Google Sheet access is required to get started, even though the supported contributor path is actually:

- create the local schema
- sync the public OpenCRE graph from upstream
- optionally import local mappings via MyOpenCRE / CSV import

That documentation gap is what issue `#599` is about.

## Solution
This PR clarifies the supported contributor workflow in the docs by:

- explicitly stating that Google Sheet access is not required for ordinary contribution
- documenting `make migrate-upgrade` followed by `make upstream-sync` as the standard local bootstrap path
- explaining that `make upstream-sync` pulls the public OpenCRE graph into the local database
- documenting the optional local import path using `CRE_ALLOW_IMPORT=true`, MyOpenCRE, and the CSV import endpoints
- reframing spreadsheet-based commands as optional / specialized rather than required for normal development

## Changed Files

Changed files:

- `README.md`
- `docs/developmentSetup.md`

This is a docs-only change. It does not modify runtime behavior, import logic, or database code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,16 @@ Install dependencies:
 make install
 ```
 
-Download the latest CRE graph from upstream by running:
+Create the local schema and download the latest CRE graph from upstream:
 
 ```bash
+make migrate-upgrade
 make upstream-sync
 ```
+
+For contributors, this is the supported local data bootstrap path. You do
+**not** need access to the internal OpenCRE Google Sheet to work on the
+project.
 
 You can precompute local gap-analysis cache after imports with:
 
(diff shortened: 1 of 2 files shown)


northdpole (changes requested):
Thanks @Bornunique911 — this closes the #599 docs gap cleanly.

Requesting two small follow-ups:
1. Remove the `--review --from_spreadsheet` example from README — `--review` is currently commented out in `cre.py`, so that CLI invocation fails. Keep `--add --from_spreadsheet` and mark spreadsheet/gspread as maintainer-only.
2. Note that `POST /rest/v1/cre_csv_import` expects multipart field `cre_csv` (as in the API).

Otherwise LGTM for a docs-only merge.

northdpole (approved):
Thanks @Bornunique911 — both follow-ups look good (`--review` removed, `cre_csv` field + curl example documented, spreadsheet path marked maintainer-only).

Approving for merge. This closes the #599 docs gap.

PR #1005, opened 2026-08-04
feat(frontend): resource-selection UI in MyOpenCRE (Part of #586)

skypank-coder:
## What & why
Final core piece of #586, building on the merged backend chain: user + selection
persistence (#980), the GET/PUT /rest/v1/user/resources API (#981), and server-side
filtering of /rest/v1/standards (#1001). This wires the frontend on top.
Frontend-only — no backend changes.

## Changes
- **`useResourceSelection` hook** — GET/PUT `/rest/v1/user/resources`, same
  raw-fetch/status handling as `useUser` (200 ok, 401 → anonymous, other → error);
  returns `{ selected, loading, save, saving, error }`.
- **`ResourceSelector` component** — fetches the full universe via
  `/rest/v1/standards?all=true` (must bypass the per-user filter, or the picker
  could never show unselected standards), renders a semantic-ui-react checkbox
  list pre-checked from the saved selection, with an explicit **Save** button and
  Loader / saved / error `<Message>` states. Prompts to log in when logged out;
  renders nothing when the MyOpenCRE capability is off.
- **MyOpenCRE page** — new "Your standards" section (no new route/nav; reuses the
  existing `capabilities.myopencre` gating).

Reuses `useUser` / `useCapabilities` / `useEnvironment` unchanged. react-router v5
+ semantic-ui-react only. No backend, migration, or auth-route changes.

## Tests / first jsdom component-test lane
This adds the repo's **first jsdom component-test lane**: `jest.component.config.js`
(+ `src/setupTests.ts`) and a new `yarn test` script, reusing `@testing-library/react`
+ `ts-jest` already in `package.json`. The existing `test:e2e` / puppeteer config is
**untouched** (no `jest.config.js` is introduced, so bare `jest` keeps its defaults),
and this lane is **not** wired into `make frontend` or CI yet.

10 tests: hook load/401/save-PUT-body; component `?all=true` fetch, pre-checked
selection, toggle+Save, success + error messages, logged-out prompt, myopencre-off gate.

## Verification (local)
- `yarn test` → 10 passed
- `yarn test:e2e --listTests` still lists `basic-e2e.test.ts` (e2e config unchanged)
- `make frontend` (webpack prod) builds clean
- `yarn lint` (prettier) applied

PR #1008, opened 2026-08-08
fix(db): concurrency-safe set_user_resource_selection + enforce OpenCRE (Part of #586)

skypank-coder:
## What & why
The small follow-up @northdpole asked for on the #586 resource-selection work.
Two things, both scoped to the selection write path:

1. **Concurrency recovery in `set_user_resource_selection`.** The method does a
   delete-then-insert; two concurrent PUTs for the same user could both delete
   then insert the same `(user_id, standard_name)`, and the second commit raised
   `IntegrityError` on `uq_user_resource_selection` → a 500. It now rolls back and
   retries the replace **exactly once** (after the competing txn has committed,
   the retry deletes its rows and inserts cleanly). A second failure propagates —
   no retry loop. Mirrors the existing `upsert_user` recovery pattern.

2. **"OpenCRE is always included" enforced at write.** A non-empty selection now
   always persists `OPENCRE_STANDARD_NAME`, so `GET /user/resources` reflects the
   invariant the read filter and UI copy already assume. Done in the PUT handler
   (API layer) — not in `db.py`, to avoid a circular import via `gap_analysis`.
   **An empty PUT stays `[]`** (which the `/standards` filter treats as "show
   everything"); injecting OpenCRE there would wrongly narrow it to "OpenCRE only".

## Scope
`db.py` (recovery) + `web_main.py` PUT handler (OpenCRE) + tests. No migration,
no model change, no frontend, no auth-route change.

## Tests
- Deterministic recovery: mock the commit to raise `IntegrityError` once → the
  method rolls back, retries once, returns the correct deduped selection.
- Persistent failure re-raises (retry is bounded to one — no infinite loop).
- OpenCRE: non-empty PUT injects it, PUT with it present doesn't duplicate, and
  **empty PUT stays empty**. Existing exact-list PUT tests updated for the invariant.
- The real Postgres `uq_user_resource_selection` constraint is exercised by the
  existing unique-constraint test.

## Verification
Verified on real Postgres (constraint-driven): 30 tests green incl. the recovery
and OpenCRE cases; `black` clean; no new mypy errors.

### DIFF
--- a/application/database/db.py
+++ b/application/database/db.py
@@ -1353,7 +1353,16 @@ class Node_collection:
             if name not in deduped:
                 deduped.append(name)
 
-        def _replace() -> None:
+        def _replace() -> List[str]:
+            # Serialize concurrent replacements for this user by locking the
+            # parent User row: a second PUT for the same user waits here until
+            # ours commits, then its DELETE sees our rows. Without this, two
+            # *disjoint* selections could both commit (no IntegrityError) and
+            # merge, breaking replacement semantics. FOR UPDATE is a no-op on
+            # SQLite (dev/tests).
+            self.session.query(User).filter(
+                User.id == user_id
+            ).with_for_update().first()
             self.session.query(UserResourceSelection).filter(
                 UserResourceSelection.user_id == user_id
             ).delete()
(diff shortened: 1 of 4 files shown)


northdpole (approved):
## Review — concurrency-safe resource selection (#1008 / #586 follow-up)

Looks good. Addresses the post-#1005 ask cleanly.

### What works
- **Per-user serialization:** `User` row `FOR UPDATE` before delete/insert so two disjoint concurrent PUTs cannot merge selections. Documented SQLite no-op is fine for local/tests.
- **IntegrityError recovery:** rollback + exactly one retry; persistent failure re-raises. Matches `upsert_user` style.
- **Return under lock:** `get_user_resource_selection` before `commit` so the response cannot race a later writer.
- **OpenCRE injection:** non-empty selections always include `OPENCRE_STANDARD_NAME`; empty `[]` stays empty (“show all”) — correct, and covered by API tests.
- **Tests:** flaky-commit retry, persistent IntegrityError, Postgres lock-blocking test (skip on SQLite), OpenCRE inject/dedupe/empty cases.

CodeRabbit’s serialize note is addressed by the row lock; BLE001 noqa on the worker is justified.

**Approve.** ~11 behind `main` — light rebase optional before merge; unlikely to conflict on these files. Happy to rebase-merge when you want it landed.

northdpole:
Merged (rebase) after updating onto `main` — CI green. Thanks @skypank-coder — looking forward to the #963 auth-routes PR when you're ready.

PR #1010, opened 2026-08-10
Fix flaky gap-analysis tests connecting to real Redis instead of mock

### LINKED ISSUE #1009: Gap-analysis tests silently connect to real Redis instead of the mock, causing flaky/order-dependent failures
## Description

Several tests in `application/tests/web_main_test.py` that cover the
`/rest/v1/map_analysis` endpoint mock the Redis connection (`redis.from_url`)
and job queueing (`rq.Queue.enqueue_call`), but `application/utils/redis.py`'s
`connect()` function has a code path that bypasses both mocks whenever
`REDIS_HOST` and `REDIS_PORT` are set — which they are by default for anyone
following this project's own documented local setup (`.env.example` sets
`REDIS_HOST=localhost` / `REDIS_PORT=6379`, and `cre.py` calls
`load_dotenv()` unconditionally at import time). This causes these tests to
silently connect to a real local Redis instance instead of the mock,
producing flaky, order-dependent, environment-dependent failures.

## Demo

Screen recording showing the actual bypass happening in real time: the
Redis key is empty, the "mocked" test runs, and the key now exists in real
Redis with a real ~36-hour TTL — proving the test wrote to a real server
despite being fully mocked on paper.

https://github.com/user-attachments/assets/bc31bd83-9dc5-4cf8-86b2-42d3ebb6c813

## Steps to reproduce

Ensure a local Redis is reachable at `localhost:6379` (e.g. via
`make start-containers`), and that `.env` has `REDIS_HOST=localhost` and
`REDIS_PORT=6379` set (the documented default from `.env.example`).

Step 1 — confirm the target key doesn't exist yet:

`docker exec <redis-container> redis-cli GET "ga:inflight:aaa >> bbb"`

Result: nothing (key absent).

Step 2 — run the test:

`python -m unittest application.tests.web_main_test.TestMain.test_gap_analysis_create_job_id -v`

Step 3 — check the same key again immediately after:

`docker exec <redis-container> redis-cli GET "ga:inflight:aaa >> bbb"`

`docker exec <redis-container> redis-cli TTL "ga:inflight:aaa >> bbb"`

Result: now returns a real value with a real ~36-hour TTL, proving the test
wrote directly to real Redis via `conn.set(...)` / `conn.expire(...)` in
`application/web/web_main.py` (`map_analysis()`), regardless of whether the
test's own assertion happened to pass or fail that run.

**Definitive root-cause isolation** — the same test flips outcome purely
based on this one condition. With Redis env vars blocked from loading:

`$env:REDIS_HOST=""; $env:REDIS_PORT=""`
`python -m unittest application.tests.web_main_test.TestMain.test_gap_analysis_create_job_id -v`

Result: OK.

With `.env`'s real values loaded (the default/documented setup):

`python -m unittest application.tests.web_main_test.TestMain.test_gap_analysis_create_job_id -v`

Result: FAILS intermittently depending on what's cached in real Redis.

## Expected behavior

These tests should be fully isolated from any real Redis instance
regardless of the local environment/`.env` configuration — mocking
`redis.from_url` should be sufficient (or the correct branch of
`connect()` should also be mocked/patched).

## Actual behavior

`application/utils/redis.py`'s `connect()` has three branches:

```python
def connect():
    redis_url = os.getenv("REDIS_URL", "redis://localhost:6379")
    if os.getenv("REDIS_HOST") and os.getenv("REDIS_PORT"):
        return redis.StrictRedis(host=..., port=..., ...)   # never mocked by the tests
    elif redis_url:
        if redis_url == "redis://localhost:6379":
            return redis.from_url(redis_url)                 # this is the only branch the tests mock
        else:
            return redis.Redis(...)
    else:
        ...

DevPatils:
## Summary

Fixes #1009 — several tests in `web_main_test.py` covering
`/rest/v1/map_analysis` were silently connecting to a real local Redis
instance instead of the mock, causing flaky, order-dependent,
environment-dependent test failures.

## Root cause

The tests patched `redis.from_url` (the third-party library's URL-based
connection factory), but `application/utils/redis.py`'s `connect()`
function has three possible branches and takes a **different, unmocked**
one — `redis.StrictRedis(host=..., port=...)` — whenever `REDIS_HOST` and
`REDIS_PORT` environment variables are set. These are set by default for
anyone following this project's own documented local setup
(`.env.example` sets `REDIS_HOST=localhost` / `REDIS_PORT=6379`, loaded
automatically via `load_dotenv()` in `cre.py`). So the standard,
documented local dev setup silently defeated the mock for anyone running
these tests.

## The fix

Patch `application.utils.redis.connect` directly, at its actual point of
use in `web_main.py`, instead of patching one specific branch
(`redis.from_url`) of its internal implementation. This makes the tests
immune to which branch `connect()` takes based on environment variables —
there's now exactly one thing to mock, and it's the thing that's actually
called.

```python
# before
@patch.object(redis, "from_url")

# after
@patch.object(app_redis, "connect")

### DIFF
--- a/application/tests/web_main_test.py
+++ b/application/tests/web_main_test.py
@@ -10,7 +10,6 @@ import tempfile
 from types import SimpleNamespace
 from unittest.mock import patch, Mock
 
-import redis
 import rq
 import os
 import networkx as nx
@@ -20,6 +19,7 @@ from application.tests.utils import data_gen
 from application.database import db
 from application.cmd import cre_main
 from application.utils import spreadsheet
+from application.utils import redis as app_redis
 from application.defs import cre_defs as defs
 from application.web import web_main
 from application.utils.gap_analysis import GAP_ANALYSIS_TIMEOUT
@@ -741,7 +741,7 @@ class TestMain(unittest.TestCase):
             self.assertEqual(404, response.status_code)
             mock_redirect.assert_called_once_with("CWE", "999")
 
-    @patch.object(redis, "from_url")
+    @patch.object(app_redis, "connect")
     @patch.object(db, "Node_collection")
     def test_gap_analysis_from_cache_full_response(
         self, db_mock, redis_conn_mock
@@ -762,7 +762,7 @@ class TestMain(unittest.TestCase):
     @patch.object(db, "Node_collection")
     @patch.object(rq.job.Job, "fetch")
     @patch.object(rq.Queue, "enqueue_call")
-    @patch.object(redis, "from_url")
+    @patch.object(app_redis, "connect")
     def test_gap_analysis_from_cache_job_id(
         self, redis_conn_mock, enqueue_call_mock, fetch_mock, db_mock
     ) -> None:
@@ -784,7 +784,7 @@ class TestMain(unittest.TestCase):
     @patch.object(db, "Node_collection")
     @patch.object(rq.job.Job, "fetch")
     @patch.object(rq.Queue, "enqueue_call")
-    @patch.object(redis, "from_url")
+    @patch.object(app_redis, "connect")
     def test_gap_analysis_returns_existing_inflight_job(
         self, redis_conn_mock, enqueue_call_mock, fetch_mock, db_mock
     ) -> None:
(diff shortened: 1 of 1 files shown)


DevPatils:
@northdpole can you please review it once?

northdpole:
Reviewed — correct fix. Patching `application.utils.redis.connect` instead of `redis.from_url` covers all connection branches (StrictRedis when REDIS_HOST/PORT are set). LGTM, merging.

PR #1012, opened 2026-08-12
week_8b: Module C (The Librarian) — docs, final metrics, and the regression gate

PRAteek-singHWY:
Hi @northdpole — Week 8b of Module C, and the last piece of the plan that is not code: package docs, the runbook, measured final metrics, the GSoC final report, and the CI gate.

**Stacked on #1011 (week_8).** Until that merges the diff shows its commits too; I'll rebase onto `main` as it lands, shrinking this to the docs-only surface. Stack order: **#1011 → this**.

No application code changes. One `.gitignore` edit, one workflow, four markdown files.

## Overview

Module C is complete through C.4 and now drains B's live queue (#1011). What was missing was everything a maintainer needs to *operate* and *trust* it: how to run it, what each knob means, what the numbers actually are, and a gate that stops those numbers rotting.

## What changed

| Area | Files | Description |
|---|---|---|
| Package README | `application/utils/librarian/README.md` | The C.-1 → C.4 stages, the supporting live-path modules, the two invariants (consumption gated on persistence; errors ≠ refusals), and the design constraints — seams not implementations, no DB import at module scope, declared-degraded over silently-degraded. |
| Runbook | `docs/gsoc_2026_module_c/runbook.md` | Three ways to run C, every CLI flag, all 8 `CRE_LIBRARIAN_*` variables, the pgvector cache migration, and a troubleshooting section keyed to the actual error strings. |
| Final metrics | `docs/gsoc_2026_module_c/final_metrics.md` | Every number reproducible from one command, with the environment stated. |
| Final report | `docs/gsoc_2026_module_c/final_report.md` | The GSoC write-up: what was built week by week, design decisions worth defending, results, integration status, what is not built, and recommendations. |
| Regression gate | `.github/workflows/librarian_regression.yml` | Hermetic CI: the librarian unit suite plus the golden-set decision gate, path-filtered to Module C. |
| gitignore | `.gitignore` | `*.md` is blanket-ignored, so the new docs need explicit negations — the same pattern `docs/Mid_eval_blog_gsoc2026/module_B_mideval_blog.md` already uses. Nothing else is un-ignored. |

## The measured numbers

| Stage | Result | Target | |
|---|---|---|---|
| C.0 boundary validation | 319/319 (100%) | — | |
| C.0.5 explicit resolver | 5/5 (100%) | 100% | ✅ |
| C.1 retrieval recall@20 | 285/292 (98%) any-hit | — | |
| C.2 rerank top-1 | 220/292 (75%) | ≥ 90% | ❌ |
| C.3 calibration ECE | 0.046 (fitted T=1.105) | < 0.10 | ✅ |
| C.4 review recall | **5/5 (100%)** | — | ✅ |
| C.4 auto-link recall | 176/314 (56%) | — | |

**Read C.4 by direction.** Review recall is 5/5 — every chunk that should reach a human does, and the engine never wrongly auto-links something that needed review. For a gate whose job is protecting a shared graph, that is the number that matters. Auto-link recall of 56% is the cost: at τ=0.80 many correct-but-close positives fall under the bar and route to review, which is the safe direction.

**On the missed top-1 target, stated plainly rather than rounded away.** Week 8's investigation tried thirteen reranker levers and all thirteen regressed — C.2 as shipped scores net −7 against plain cosine on the same shortlist. The cause is not the model: **427 of 428 CREs have empty `description` fields**, and a cross-encoder cannot cross-attend to a bare title. C.1 still reaches 98% recall over that same corpus, which localises the problem precisely — the information needed to *rank within* a shortlist largely is not present. The path to 90% runs through populating CRE descriptions, and that is a corpus problem upstream of Module C.

W9's selective reranking (250/319 top-1 vs the shipped 238, precision held at τ=0.80) is documented as measured but **deliberately unshipped**: it needs two calibrators and held-out validation. Shipping it on in-sample numbers would be the same greenwashing this module spent four separate fixes removing.

## Why the CI gate is hermetic

No DB, key, or model download. The semantic reports need live CRE vectors, and seeding a candidate pool from golden text offline is exactly the leakage the hub firewall exists to strip — a CI job that "measured" C.2 offline would be measuring nothing.

What it *does* catch is the failure that actually bit #1011: a Module B schema change silently rejecting every row at the C.0 boundary, leaving the explicit gate with nothing to count while the run still exited 0.

## What the report records as not built

- Graph / review-queue writers (W8b) — C emits envelopes; no link is committed.
- The SafetyGuard detector — the seam is wired, the detector is future work.
- Selective reranking — measured, promising, unvalidated.

It also records that **A → B is not yet connected** (Module A does not write `harvest_input`), so the end-to-end chain is not demonstrable outside a seeded record. That belongs in the report rather than in a claim that the pipeline is whole.

## How to verify locally

```bash
python -m unittest discover -s application/tests/librarian -p '*_test.py' -t .
python scripts/evaluate_librarian.py \
    --dataset application/tests/librarian/fixtures/golden_dataset.json

# the live numbers in final_metrics.md (needs a populated cache + embedding LLM)
python scripts/rewrite_sqlite_embeddings_to_vec.py --db standards_cache.sqlite
python scripts/evaluate_librarian.py \
    --dataset application/tests/librarian/fixtures/golden_dataset.json \
    --use_live_embeddings --cache_file standards_cache.sqlite
```


PRAteek-singHWY:
Folding this into #1011 rather than reviewing it as a stack.

The split was mine and it was not worth the extra review round: this is one week's work, and separating the integration from its own docs meant two PRs, two CI runs and two review cycles for a change that reads better in one place. #1011 now carries all 11 commits — the live B→C integration plus the package README, runbook, final metrics, final report, and the regression-gate workflow.

Nothing is dropped; #1011 is a strict superset of what was here. Closing.

Sorry for the churn, @northdpole — please review #1011 only.

PR #1014, opened 2026-08-12
feat(workstream-e): LLM re-rank + LangGraph decision flow for cheatsh…

shreeshtripurwarcomp23-coder:
## Summary

Implements RFC Workstream E — LLM Re-Rank and Decision Graph (LangGraph) —
from `docs/rfc/cheatsheets-llm-autonomous-mapping-rfc.md` (Issue E, section 12).

Given a `CheatsheetRecord` (Workstream B) and a shortlist of `CandidateCRE`
(the contract Workstream D's `retrieve_candidate_cres` is expected to
return), `rerank_candidates_with_llm()` runs a small LangGraph flow that:

1. Sends the cheat sheet's title/summary/headings and the candidate CREs to
   an LLM, asking it to score (0-1) and justify each candidate.
2. Validates the response against a strict Pydantic schema and drops any
   `cre_id` the model invents that wasn't in the original shortlist.
3. Falls back to retrieval-only scoring — deterministically, per candidate —
   on LLM error, timeout (hard wall-clock cutoff, default 30s), or malformed
   output, so a cheat sheet can never be dropped from the pipeline by an LLM
   hiccup.
4. Buckets every result into `high` / `medium` / `low` confidence via
   `classify_confidence()`, using the RFC's bootstrap thresholds
   (`>= 0.85` / `>= 0.70`), overridable by env var for later recalibration.
5. Attaches an audit `RerankTrace` (model, prompt version, UTC timestamp,
   fallback flag/reason) to every result.

## Why `CandidateCRE` is defined here

Workstream D (`retrieve_candidate_cres`) hasn't landed yet. `CandidateCRE`
(`cre_id`, `score`, `text`) mirrors the RFC's contract exactly, so wiring in
the real Workstream D output later is a drop-in swap.

## Design choices worth flagging for review

- LLM call is dependency-injected (`llm_score_fn`), matching the seam
  pattern already used in `embed_alignment.py` (`ai_client`) and
  `librarian/cross_encoder.py` (`score_fn`). Production defaults to a
  lazily-imported LiteLLM call; the entire test suite runs with a stub,
  no network/API key required.
- Added `langgraph` to `requirements.txt` (`>=0.2,<1`).
- Did not touch `prompt_client.py` — kept this fully decoupled.

## Note for maintainers

`application/utils/librarian/` (already on main) implements a similar
retrieve → rerank → decide → emit flow for a generic `knowledge_queue`,
using a cross-encoder rather than an LLM. This PR is scoped strictly to
the RFC's Workstream E contract for cheat sheets, as assigned. Happy to
align the two if that's useful.

## Testing

16 new tests in `application/tests/cheatsheet_rerank_test.py`, all passing,
alongside the existing cheatsheet test suite. black/flake8 clean.

### DIFF
--- a/application/tests/cheatsheet_rerank_test.py
+++ b/application/tests/cheatsheet_rerank_test.py
@@ -1,13 +1,11 @@
 import time
 import unittest
 
-from application.defs.cheatsheet_defs import CheatsheetRecord
 from application.utils.external_project_parsers.parsers.cheatsheet_rerank import (
-    CandidateCRE,
     RerankError,
-    build_rerank_graph,
+    build_rationale_graph,
     classify_confidence,
-    rerank_candidates_with_llm,
+    generate_link_rationale,
 )
 
 # LangGraph's first StateGraph().compile() in a process pays a one-time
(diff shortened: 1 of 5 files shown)


shreeshtripurwarcomp23-coder:
@robvanderveer @Pa04rth I'd love a review on this, thanks!

northdpole (changes requested):
## Review — Workstream E (#1014)

Not merging this as-is.

**Blocker — new production dependency:** `requirements.txt` adds `langgraph>=1.0.10,<2`. That is a Heroku-slug and supply-chain decision, not a drive-by import. Please do not land LangGraph on the production web slug unless we agree it belongs there.

Options that would unblock:
1. Keep the graph as an **optional extra** (`requirements-dev.txt` or `requirements-gsoc.txt`) and lazy-import `langgraph` so the web app does not install it, **or**
2. Implement the same rerank/classify/fallback flow without LangGraph (plain functions + tests), and leave the RFC as design notes.

Also please confirm this is not on the production import path yet (parser under `external_project_parsers` with no CLI/cre.py wiring looks unused — good for a checkpoint, bad if `requirements.txt` still grows).

CI otherwise looks green. Happy to re-review a version without the prod dep.

ParthAggarwal16:
yo, randomly hopped on this and noticed something, i dont think pool.shutdown(wait=False) can cancel a running thread? or am i missing something regarding this?

PR #1024, opened 2026-08-18
feat: cheat sheet suggestion → ParseResult import adapter (Workstream F, F3)

skypank-coder:
## What & why
Workstream F checkpoint F3, building on the merged F1–F2 contract. Adds
`suggestions_to_parse_result(approved, cache)` — converts approved MappingSuggestions
into a `ParseResult` the existing OpenCRE import flow accepts. This is the adapter
between the review artifacts and the register/import path.

## What it does
Per approved suggestion: builds `defs.Standard(name="OWASP Cheat Sheets", section=title,
hyperlink, tags=build_tags(GUIDANCE/CHEATSHEET/DEVELOPER/STABLE, source="owasp_cheatsheets",
category as extra tag))`, resolves each candidate `cre_id` via `cache.get_CREs`, and adds
`AutomaticallyLinkedTo` links for resolved CREs. Unknown cre_ids are logged and skipped;
zero-link Standards are dropped; advisory fields (score/confidence/reason/cheatsheet_id)
are not persisted.

## Real-code guards (matched against the live cheatsheets parser)
- `shallow_copy()` on each linked CRE — mirrors the supplemental-cheatsheets path, strips
  the CRE's own hydrated links.
- `has_link` guard — two candidates resolving to the same CRE collapse to one link rather
  than raising `DuplicateLinkException`.
- Empty `category` → `extra=[]`, so no blank classification tag is emitted.

## Tests
15/15 green (7 new F3 + 8 existing F1–F2), no Postgres/Neo4j — a stub cache replaces the
DB. Covers: link building, `validate_classification_tags` passes on the output (the real
import-flow gate), unknown-id skip with siblings still linked, all-unknown → Standard
omitted, deterministic output, the duplicate-cre_id guard, and the empty-category case.

## Scope
F3 adapter only. F4/F5 (CLI + end-to-end wiring) are a follow-up. No F1–F2 or web-app
changes.

### DIFF
--- a/application/tests/cheatsheets_workstream_f_test.py
+++ b/application/tests/cheatsheets_workstream_f_test.py
@@ -1,12 +1,14 @@
-"""Tests for Cheat Sheet -> CRE mapping, Workstream F, checkpoints F1+F2.
+"""Tests for Cheat Sheet -> CRE mapping, Workstream F, checkpoints F1+F2+F3.
 
 F1: the suggestions.json data contract -- SUGGESTIONS_SCHEMA plus the
 CandidateCRE / MappingSuggestion dataclasses.
 F2: the read/write adapters -- write_suggestions_json and
 load_approved_suggestions (schema-validate -> parse -> filter approved).
+F3: the import adapter -- suggestions_to_parse_result (approved suggestions ->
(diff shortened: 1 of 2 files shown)


Pa04rth:
Please update the branch , then will merge

Pa04rth (approved):

PR #1034, opened 2026-08-21
feat(module-b): retry infrastructure-failed chunks instead of pinning them UNCERTAIN

manshusainishab:
## Problem

Module B's Stage-2 classifier is "mark-and-continue": when an LLM call fails (rate-limit exhaustion, network, auth), the affected chunks are written to `knowledge_queue` as `UNCERTAIN` with `confidence = 0.0`, and their `harvest_input` rows are marked `processed`. Because the queue dedups on `content_hash` (`ON CONFLICT DO NOTHING`), that conf-0.0 row is then **pinned forever** — a later run can neither re-read the input (no longer `pending`) nor replace the row. A transient outage permanently degrades a possibly-`KNOWLEDGE` chunk to `UNCERTAIN 0.0`; a whole-run outage produces an all-`UNCERTAIN` run that still exits `0`, so the orchestrator has no signal to retry.

Not a data-loss bug (recall-first still holds — the chunk reaches Module C as `UNCERTAIN`), but a real robustness gap: B treats an **infrastructure failure** (should be retried) the same as a **genuine `UNCERTAIN` verdict** (should be persisted).

## Fix

The classifier now distinguishes the two (`is_infra_failure`): only the LLM-call-exception path (`llm_call_failed`) is retryable; unparseable model output (`malformed_output`) stays a persisted `UNCERTAIN` (retrying junk would loop).

- **Leave infra-failed chunks `pending`.** They are not written, and their `harvest_input` rows are left `pending` for a natural retry on a subsequent run. Surfaced as a new `retry_pending` field in the JSON summary. The successfully-classified rows still commit (idempotent).
- **Signal the orchestrator on a total wipeout.** `status="degraded"` and a non-zero exit when `retry_pending / classified >= CRE_NOISE_FILTER_FAILURE_THRESHOLD` (default **1.0** — only when *every* classified chunk failed). A partial failure stays `ok` (exit 0) with `retry_pending > 0` for the orchestrator to drain.

## Verification

- **+6 tests → 103 noise_filter tests green**; `black --check` clean.
- **Live Postgres (Docker pgvector, full `flask db upgrade`):**
  - normal run → `retry_pending 0, status ok`, exit 0, rows `processed`, keepers queued;
  - forced outage (bogus model) → `retry_pending N, status degraded`, **exit 1**, rows stay `pending`, nothing written;
  - re-run same `run_id` with a working model → the `pending` rows drain, exit 0.
- **Classification unchanged:** the labeled-set eval still scores **0.930 accuracy / KNOWLEDGE recall 1.000 / 0 leakage** (identical to the mid-eval); this change only alters infra-failure handling.

Docs: `module_b_runbook.md` updated (§1 env var, §4 summary `retry_pending` + `degraded`, §7 guarantees).

### DIFF
--- a/application/cmd/cre_main.py
+++ b/application/cmd/cre_main.py
@@ -893,6 +893,8 @@ def run(args: argparse.Namespace) -> None:  # pragma: no cover
         return
 
     if getattr(args, "run_noise_filter", False):
+        import sys
+
         from application import sqla
         from application.utils.noise_filter.pipeline import run_noise_filter
 
@@ -904,6 +906,12 @@ def run(args: argparse.Namespace) -> None:  # pragma: no cover
             dry_run=getattr(args, "noise_filter_dry_run", False),
         )
         print(summary.to_json())
+        if summary.status == "degraded":
+            # A whole run of infrastructure failures. The classified rows are
+            # already committed (idempotent) and the failed ones stay `pending`,
+            # so exit non-zero to have the orchestrator retry -- a retry then
+            # re-processes only the pending rows.
+            sys.exit(1)
         return
 
     if args.add and getattr(args, "from_ai_exchange_csv", None):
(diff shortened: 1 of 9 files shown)


manshusainishab:
@coderabbitai can you review this PR.

Pa04rth (approved):

PR #1038, opened 2026-08-22
GSoC Module_A-week7: deduplication and metrics (stacked on top of #1029)

ParthAggarwal16:
Summary

Implements the Week 7 deduplication stage for the harvester pipeline.

Changes

* Add deduplication logic for previously processed/duplicate content.
* Integrate deduplication with the existing harvester pipeline.
* Add supporting models and tests.
* Ensure duplicate artifacts are filtered before downstream processing.

Testing

* Added/updated unit tests covering deduplication behavior and pipeline integration.

1. High-level architecture
image

2. Repository synchronization flow
image

4. Incremental processing sequence
image

5. Document and deduplication model
image

6. Deduplication state diagram

image

7. Persistence diagram

image

8. Configuration validation flow
image

smoke tests:

image
image

image
image

image
image

PR #1044, opened 2026-08-24
GSoC Module_A-week8: feat(harvester): chunking retrieval (stacked on top of #1038 )

ParthAggarwal16:
Week 8: Semantic Document Chunking Pipeline
Summary

Adds a document chunking foundation to the harvester pipeline: documents produced by earlier pipeline stages are now split into semantically coherent chunks, enriched with heading-path/line-range metadata, converted into RFC-facing ingestion records, and validated before being returned. Also includes CI stability fixes and dependency cleanup surfaced during review.

What's new

Semantic chunking (chunker.py)

DocumentChunker wraps LlamaIndex's SemanticSplitterNodeParser + HuggingFaceEmbedding (sentence-transformers/all-MiniLM-L6-v2 by default) to split document text into semantically coherent segments.
Returns ChunkInfo objects (text, start_char_idx, end_char_idx) preserving exact offsets into the source text.
Empty/whitespace-only documents short-circuit to [] without invoking the splitter.

Structure-aware chunk records (chunk_record_builder.py)

ChunkRecordBuilder converts ChunkInfo objects into IngestChunkRecords, resolving:
the active Markdown heading path for each chunk (based on the chunk's starting line and the document's heading_structure)
1-based inclusive line ranges derived from character offsets
deterministic chunk_ids built from artifact ID, heading path, char offsets, and a content hash — so identical text at different offsets gets distinct IDs, while rebuilding the same chunk twice is idempotent
span.index / span.total for chunk ordering

Validation (chunk_record_validator.py)

ChunkRecordValidator enforces non-empty text/IDs, valid chunk_id prefix, consistent span.index/span.total, valid char and line ranges before a record is considered usable.

Pipeline glue (chunk_pipeline.py)

DocumentChunkPipeline composes chunker → record builder → validator into a single chunk(document) call, validating every record before returning them (fixed in the CodeRabbit follow-up — the pipeline previously skipped validation, so invalid records from a swapped-in chunker/builder could bypass it).

Benchmark

chunking_benchmark_test.py added as an opt-in benchmark (RUN_CHUNKING_BENCHMARK=1), consistent with the existing diff-pipeline benchmark pattern, so it doesn't hit HuggingFace/network on every CI run.

Testing
chunker_test.py, chunk_record_builder_test.py, chunk_record_validator_test.py, chunk_pipeline_test.py — new unit tests covering empty input, node-boundary/order preservation, heading-path resolution, chunk ID determinism/uniqueness, and validator rejection paths.
chunking_benchmark_test.py — opt-in perf smoke test, skipped by default.
Full harvester test suite passes locally (pytest application/tests/harvester_test/).

High-level architecture
image

Data-flow diagram
image

Chunk processing sequence
image

Domain model diagram
image

Current boundary diagram
image

smoke tests:

image

image

image

image

image

image

image

image

### DIFF
--- a/application/tests/harvester_test/artifact_registry_test.py
+++ b/application/tests/harvester_test/artifact_registry_test.py
@@ -1,0 +1,99 @@
+import unittest
+from datetime import datetime
+
+from application.utils.harvester.artifact_registry import ArtifactRegistry
+from application.utils.harvester.models import ArtifactRegistryRecord
+
+
+class ArtifactRegistryTests(unittest.TestCase):
+    def test_insert_record(self):
+        registry = ArtifactRegistry()
+
+        record = ArtifactRegistryRecord(
+            artifact_id="art:test:file.md",
+            repository="OWASP/ASVS",
+            locator_path="file.md",
+            content_hash="abc",
+            last_commit_sha="123",
+            last_pipeline_run="run1",
+            last_processed_at=datetime.now(),
+            status="new",
+        )
+
+        registry.upsert(record)
+
+        self.assertTrue(registry.exists(record.artifact_id))
+
+    def test_get_record(self):
+        registry = ArtifactRegistry()
+
+        record = ArtifactRegistryRecord(
+            artifact_id="art:test:file.md",
+            repository="OWASP/ASVS",
+            locator_path="file.md",
+            content_hash="abc",
+            last_commit_sha="123",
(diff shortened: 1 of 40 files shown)


northdpole:
Superseded by mod-a-nits (includes week 8 chunking rewritten without LlamaIndex on prod requirements, plus harvest_input CLI).

Issue #1021, opened 2026-08-16
OAuth /rest/v1/callback continues after Flask session state mismatch (missing return)


shiwani42:
### **What is the issue?**

In `callback()`, oauthlib `MismatchingStateError` returns a redirect. The Flask session check does not:

```python
if not session.get("state") or session.get("state") != request.args["state"]:
    redirect(url_for("web.login"))  # State does not match!
credentials = flow_instance.flow.credentials
# verify_oauth2_token + session["google_id"] still run
```

This is the cookie binding for “this browser started this login.” `fetch_token` still requires a valid Google code matching the Flow singleton. A forged `state` alone does not skip Google.

### Expected Behaviour

Mismatched or missing Flask `session["state"]` redirects to login and does not set `google_id` / `name` / `email`.

### Actual Behaviour

After a successful `fetch_token`, a mismatched Flask state still verifies the ID token and writes the session.

### Steps to reproduce

Test client: mock `CREFlow.fetch_token` success and `id_token.verify_oauth2_token`. Set `session["state"] = "victim-state"`, request `/rest/v1/callback?state=attacker-state`. Session still contains `google_id` and `email`.

## Success criteria

- [ ] `return redirect(...)` on Flask state mismatch.
- [ ] Test: mismatch → no `google_id` / `email` in session; `verify_oauth2_token` not used to establish login.
- [ ] `make lint` / `make mypy` / `make test` green.

Ehtasham-Yasin:
Hi, I’d like to work on this issue. I’ll wait for maintainer acknowledgement before starting.

shiwani42:
Hi @Ehtasham-Yasin, thank you for your interest!

I’ve worked on this issue locally and am waiting for the maintainers’ review before creating the PR.

@northdpole @Pa04rth

northdpole:
**Maintainer triage (substance):** confirmed real bug.

In `callback()`:

```python
if not session.get("state") or session.get("state") != request.args["state"]:
    redirect(url_for("web.login"))  # missing return
```

Without `return`, execution continues into token verification / session writes after a state mismatch.

Acknowledged — open for a PR. Fix is `return redirect(...)`. Please also add a small unit/integration test if there is an existing OAuth test harness; otherwise a focused regression test is enough.

northdpole:
[...]

PR #1047, opened 2026-08-27
feat(api): migrate OAuth routes to /rest/v1/auth/* (RFC #876 TODO 1) — #963

skypank-coder:
## Summary

Migrates the OAuth/login endpoints to a canonical `/rest/v1/auth/*` namespace
(RFC #876, TODO 1 / issue #963), keeps the old paths working as
deprecation-flagged aliases, and consolidates the login gate behind a single
session predicate with **content-negotiated** unauthenticated responses.

No behavioural change for logged-in users. The visible changes are: (a) new
canonical URLs; (b) old URLs now advertise their successor via headers; and
(c) unauthenticated API/tooling calls get a clean `401` instead of being
redirected into login HTML.

### Canonical routes (new)
- `GET /rest/v1/auth/login`
- `GET /rest/v1/auth/user`
- `GET /rest/v1/auth/callback`
- `GET /rest/v1/auth/logout`

### Deprecated aliases (old paths — behaviour preserved, header-only)
`/rest/v1/login`, `/rest/v1/user`, `/rest/v1/callback`, `/rest/v1/logout` still
work and delegate to the canonical handlers, but now return:
- `Deprecation: true`
- `Link: <…canonical…>; rel="successor-version"`

They are **header-only** — no redirect to the canonical path (notably
`/callback` still completes the OAuth flow and lands on `/chatbot`), so existing
integrations don't break during the migration window.

### Single login predicate + content negotiation
- `_is_logged_in()` — the one source of truth, keyed on `session['user_id']`
  (recorded by the login flow since #980), not `google_id`/`name`.
- **`login_required` default is now `401`**, redirecting only real browsers:

  | Condition | Response |
  |---|---|
  | `NO_LOGIN=1` (dev bypass) | run the view |
  | logged in (`_is_logged_in()`) | run the view |
  | `Accept` includes `text/html` (browsers) | `302 → /rest/v1/auth/login` |
  | `Accept: application/json` | `401` |
  | `Accept: */*` (curl default) | `401` |
  | no `Accept` header | `401` |

  `text/html` is matched as a substring, so a real browser's
  `text/html,application/xhtml+xml,…;q=0.9,*/*;q=0.8` redirects while a bare
  `*/*` (curl/scripts) and `/admin/*` tooling get a machine-readable `401`
  instead of being bounced into Google login HTML.

### Frontend
`useUser`, `chatbot`, and `useResourceSelection` now call the `/auth/*` routes
and send `Accept: application/json` on their auth/API fetches, so an
unauthenticated state returns a `401` the hooks can handle rather than a `302`
the `fetch` would try (and fail) to follow.

## Review feedback addressed

- **CodeQL — "URL redirection from remote source":** removed the dead `?next=`
  from the browser auth challenge. `auth_login` never consumed a return target
  and `auth_callback` always lands on `/chatbot`, so forwarding `request.full_path`
  was a dead (and taint-flagged) value. The challenge now redirects to the
  constant `/rest/v1/auth/login`; the unused `_safe_next` helper was dropped.
  (Return-to-page is a possible future feature via a session-stored target.)
- **OAuth `redirect_uri` pointed at the deprecated route:** `CREFlow.instance`
  now builds the redirect URI from `url_for("web.auth_callback")` (canonical)
  instead of `url_for("web.callback")` (deprecated alias).
- **Broken session on persistence failure (login loop):** the OIDC callback now
  fails **explicitly** instead of redirecting as if login succeeded — it aborts
  `503` when `upsert_user` raises `SQLAlchemyError`, and `401` when the provider
  returns no `sub`. Previously these logged and redirected to `/chatbot` without
  setting `session['user_id']`, so the next `login_required` call returned `401`
  and the chatbot bounced the user back into an endless login flow.

## OpenAPI guardrail — resolved: EXEMPT

The four canonical `/rest/v1/auth/*` routes are OAuth/redirect endpoints, not
part of the documented read-only public API. They are added to
`OPENAPI_GUARDRAIL_EXEMPT_RULES` in `application/web/openapi_registry.py`,
alongside their already-exempt deprecated aliases — **no PathSpecs and no
`openapi.yaml` regeneration**. This clears the guardrail's route-coverage check
(which was failing both the "Test" and "Lint Code Base" jobs).

## ⚠️ Ops note (before retiring the deprecated `/callback` alias)
Because the OAuth `redirect_uri` now resolves to `/rest/v1/auth/callback`, that
canonical URI must be **registered in the Google OAuth console** before the old
`/rest/v1/callback` alias is removed.

## Tests
- `auth_routes_test.py` (new) — canonical routes, deprecated aliases carry the
  deprecation headers, the `user_id` predicate, the `NO_LOGIN` bypass, the full
  content-negotiation matrix (`text/html → 302`, multi-value browser `→ 302`,
  `application/json → 401`, `*/* → 401`, no-`Accept → 401`), anonymous
  `POST /rest/v1/completion → 401`, and the callback failure paths
  (persistence-failure `→ 503`, missing-`sub` `→ 401`, no-`next` redirect).
- `admin_imports_api_test.py` — added `/admin/*` `*/* → 401` and no-`Accept → 401`
  cases; existing admin tests unchanged.
- `user_resources_api_test.py` — updated to the `user_id` session predicate.

## Verification
- `black --check .` clean (24.4.2); `mypy` clean on `web_main.py`.
- OpenAPI guardrail passes (`route coverage OK`); `openapi_guardrail_test.py` green.
- The #963 suites pass. The only local failures are pre-existing/environmental
  (`litellm` not installed → those tests are skipped/installed in CI) and
  Windows-only filesystem path tests — none in the #963 blast radius.
- Net diff vs `main` is #963-only (10 files).

## Out of scope / follow-ups
- No OpenAPI PathSpecs / `openapi.yaml` changes (routes are exempt, not documented).
- No `/admin/*` route logic changes beyond the shared `login_required` default.
- Retiring the deprecated aliases (and the OAuth-console registration above) is a
  later migration step.


northdpole (approved):
Maintainer pass on the `/rest/v1/auth/*` migration.

Persistence-failure → 503 (no broken session) was already correct on this branch.
`?next` return-to-page is explicitly deferred — fine for this PR.

Pushed 13ff2897: **return** on OAuth state-mismatch redirect (same bug as #1021) + regression test. Auth route tests pass locally.

LGTM once CI is green. Closes #1021 when merged.

northdpole:
Also fixes acknowledged bug #1021 (missing `return` on state-mismatch redirect).

PR #1048, opened 2026-08-29
Fix paginated similarity lookup

IshanMaheshwari-777:
**Summary** 

Fixes pagination in the Python fallback used by the paginated embedding similarity lookups.

**Problem**

The fallback implementations could fail to process all embedding pages due to incorrect pagination boundaries and page sequencing.

In particular, the CRE lookup excluded the final page from its loop, while the fallback implementations could fetch the current page again rather than advancing to the next page. This could cause valid candidates on the final page to be missed, and the single-page case could result in the similarity loop not executing.

This was easy to miss because the fallback path is only used when `can_use_pgvector_similarity()` is false. When PostgreSQL has the required `embedding_vec` column available, similarity lookup uses the database-side pgvector implementation instead.

**Changes**

* Ensure every valid embedding page is processed exactly once.
* Fetch the next page only after processing the current page.
* Ensure the final page is processed.
* Add regression tests covering:

  * a best match existing only on the final page;
  * the single-page case;
  * both CRE and Standard/node similarity lookups.

**Testing**

* `python -m pytest application/tests/prompt_client_pgvector_similarity_test.py -v` — **8 passed**
* Strict mypy validation of the modified test file — **no errors**

The repository-wide `make mypy` target currently reports pre-existing errors in unrelated files.

CC: @northdpole


northdpole (approved):
Real pagination bug on the non-pgvector fallback path (`range(..., total_pages)` dropped the last CRE page; node path re-fetched the current page). Tests cover final-page and single-page cases. LGTM.

Optional follow-up: guard `cosine_similarity` when a page returns an empty embedding map (CodeRabbit). Not blocking.

IshanMaheshwari-777:
Went ahead and put the empty-page guard in — #1058 if you want to take a look.

Wrapped the similarity check in both _paginated methods with if embeddings:, so a page that comes back empty (e.g. a batch where every stored vector failed to parse) just gets skipped instead of getting handed to cosine_similarity, which was throwing ValueError: Found array with 0 feature(s)... on a shape mismatch. Confirmed it's a real crash by reverting the guard and reproducing that exact error, then added two regression tests (test_node_paginated_skips_empty_final_page, test_cre_paginated_skips_empty_final_page) that fail without the fix and pass with it.

Separately, I have dropped you a note on Slack about a couple of other things I came across.

PR #1052, opened 2026-09-04
fix(frontend): keep Map Analysis popup inside the viewport (#806)

### LINKED ISSUE #806: CRE information popup becomes mostly hidden in Map Analysis (Base=CWE, Compare=ASVS)
While exploring the **Map Analysis** page, I noticed a UI issue where the popup displaying **CRE information** becomes mostly hidden outside the viewport.

This occurs when using the following configuration:

**Base:** CWE  
**Compare:** ASVS  

When hovering over a mapping entry, a popup appears that shows information about the related **CRE entry** (e.g., linked CRE description and relationships). However, the popup is positioned near the right edge of the screen, causing most of its content to be hidden outside the visible area.

### Steps to reproduce
1. Open the **Map Analysis** page
2. Set **Base = CWE**
3. Set **Compare = ASVS**
4. Hover over a mapping entry to trigger the CRE information popup

### Observed behavior
The popup containing CRE information is positioned such that a large portion of it appears outside the viewport, making the content difficult to read.

### Expected behavior
The popup should remain fully visible within the viewport so that the CRE information is clearly readable.
This might involve adjusting the popup position and ensuring it stays within the visible screen area.

### Screenshot

Image

srajang1805:
Fixes #806.

## Summary
The CRE information popup on the Map Analysis page gets positioned outside the viewport when its trigger sits near the right edge of the screen, hiding most of the content.

## What changed
`application/frontend/src/pages/GapAnalysis/GapAnalysis.tsx`:
- Enable Popper.js `preventOverflow` on the CRE info `Popup` so it shifts to stay within the viewport. (semantic-ui-react already enables `flip`, but `preventOverflow` is disabled by default.)
- Cap the popup width at `min(550px, calc(100vw - 40px))` so it also fits on narrow viewports.

## How to reproduce
1. Open Map Analysis with Base = CWE, Compare = ASVS.
2. Hover over a mapping entry on the right side of the table.
3. Before: the popup extends past the right edge of the screen. After: it stays fully visible.

## Testing
- `npx prettier@2 --check application/frontend/src/pages/GapAnalysis/GapAnalysis.tsx` passes.
- `yarn build` (webpack production) compiles clean, only pre-existing bundle-size warnings.


northdpole (approved):
Simple correct fix for #806: enable Popper `preventOverflow` and cap popup width. LGTM.

PR #1056, opened 2026-09-06
Fix DiffParser dropping/misattributing non-ASCII and space-containing file paths

### LINKED ISSUE #1032: DiffParser silently drops or misattributes diffs for non-ASCII / space-containing file paths
## Description

`application/utils/harvester/diff_parser.py`'s `DiffParser.parse()` extracts changed files from a `git diff` using a regex on the `diff --git a/<path> b/<path>` header line: `match = re.match(r"diff --git a/(.+?) b/", line)` then `current_file = match.group(1) if match else None`.

This regex makes two incorrect assumptions about real-world git diff output, causing two distinct failure modes: silent data loss for non-ASCII filenames, and silent path corruption for filenames containing `" b/"`.

## Reproduction

I wrote a small, self-contained script that creates a real temporary git repository, commits real files, generates a real `git diff` from it, and runs that through the actual `DiffParser` class — no mocked/synthetic diff text, just real git output.

**Bug reproduction video:**

https://github.com/user-attachments/assets/d8c8442a-3efd-4596-80d9-4dbba8f308fa

## Expected behavior

- **Bug A:** should return 1 `DiffBlock` with `file_path='café.md'` and `added_lines=['added line']`
- **Bug B:** should return 1 `DiffBlock` with `file_path='foo b/bar.md'`

## Actual behavior

- **Bug A:** returns 0 blocks — the entire change is silently discarded, no error or warning
- **Bug B:** returns 1 block, but with `file_path` truncated to `'foo'`, silently mis-attributing the change to a nonexistent path

## Root cause

The regex in `diff_parser.py:39` assumes the header always starts with a literal `a/` and that the first `" b/"` substring found is the real separator. Neither assumption holds:

- Git C-quotes (wraps in `"..."`, escapes non-ASCII bytes) any path with non-ASCII characters by default (`core.quotePath=true` is git's default), so the header doesn't start with a literal `a/` at all in that case — the regex simply fails to match.
- The non-greedy match stops at the *first* `" b/"` substring in the line, which is ambiguous whenever the real file path itself contains that exact substring.

Existing tests in `application/tests/harvester_test/diff_parser_test.py` only cover plain ASCII, single-word filenames with no path-quoting or embedded `" b/"` — this class of edge case is completely untested.

## Impact

The harvester watches external standard repositories for changes so OpenCRE can stay in sync. A real update to a file with a non-ASCII name (plausible for translated/localized docs in an international project) would be silently missed entirely — no error, no log, just as if the change never happened. A file whose path contains `" b/"` would have its change detected but filed under the wrong document identity, corrupting downstream traceability.

## Suggested fix direction

Parse the file path from the `--- a/<path>` / `+++ b/<path>` lines instead of the `diff --git` header line — git provides these separately per side, avoiding the ambiguity of a single combined header line. Would also need proper un-quoting of git's C-quoted path format. Happy to open a PR for this.

DevPatils:
## Summary

Fixes #1032 — `DiffParser.parse()` silently discarded diffs for files with
non-ASCII names, and truncated paths containing `" b/"`, mis-attributing
those changes to a nonexistent file.

Both stemmed from a single regex on the combined header line:

`match = re.match(r"diff --git a/(.+?) b/", line)`

## What was wrong

**Bug A — silent data loss.** Git C-quotes any path containing non-ASCII
characters (`core.quotePath` is on by default), so the header for `café.md`
is `diff --git "a/caf\303\251.md" "b/caf\303\251.md"`. It starts with `"a/`,
not a literal `a/`, so the regex simply didn't match — `current_file` stayed
`None` and **the file's entire diff was dropped with no error, warning, or
log entry**.

**Bug B — silent path corruption.** The non-greedy `(.+?)` stopped at the
*first* `" b/"` substring in the line, which is ambiguous whenever the real
path contains that sequence. A file at `foo b/bar.md` produced the header
`diff --git a/foo b/bar.md b/foo b/bar.md`, and the parser reported the path
as just `'foo'`.

**Bug C — surfaced by fixing A.** A quoted `+++ "b/caf\303\251.md"` line
starts with `+`, so once headers parsed correctly it would have been
captured as *added file content*. This one never fired before only because
Bug A meant those diffs never got that far.

## The fix

Replaces the regex with explicit header parsing that:

- **Decodes git's C-quoted form** (`"caf\303\251.md"` → `café.md`) by
  undoing the octal byte escapes.
- **Uses the `a/<path> b/<path>` symmetry** to pin down the path length
  unambiguously instead of guessing which `" b/"` is the separator — outside
  of renames both sides name the same path, so the header is exactly
  `2 * len(path) + 5` characters.
- **Prefers the single-path `--- a/` / `+++ b/` headers** when present,
  since those carry one path each and are unambiguous. Falls back to the
  b-side for renames (added lines belong to the rename target).
- **Only treats `---`/`+++` as headers before the first hunk**, so `+++ x`
  *inside* a hunk is still correctly kept as the added line `++ x`.

The `--- a/`/`+++ b/` lines are preferred but not required — the existing
tests use headerless diffs (`test_multiple_files`), and those still pass
unchanged via the header path.

## Verification

Reproduced against **real `git diff` output** from actual throwaway git
repositories (not synthetic diff text), before and after:

| Case | Before | After |
|---|---|---|
| `café.md` | 0 blocks (change vanished) | 1 block, `file_path='café.md'`, `added_lines=['added line']` |
| `foo b/bar.md` | `file_path='foo'` | `file_path='foo b/bar.md'` |

- [x] Both original repro cases now pass (video/script in #1032)
- [x] 5 new regression tests added: non-ASCII paths, paths containing
      `" b/"`, quoted headers not leaking into content, in-hunk `+++`
      lines staying content, renames using the new path
- [x] Full harvester suite: 133 tests, only the 2 pre-existing
      Windows-specific path failures remain (`repository_cache_test`,
      `git_repository_client_test`) — confirmed unrelated by re-running
      them with this change stashed, where they fail identically
- [x] `black --check` passes on both modified files


DevPatils:
@northdpole please review changes once

PR #1059, opened 2026-09-08
feat(api): paginate /rest/v1/root_cres on request (#847)

### LINKED ISSUE #847: Unbounded API responses on /rest/v1/root_cres and /rest/v1/all_cres

kunalKumar-13:
Closes #847.

`/rest/v1/root_cres` returned every root CRE in one unbounded response with no way to page through it. This adds `page` and `per_page`, with the same `MAX_ITEMS_PER_PAGE` cap and the same `page` / `total_pages` keys that `/rest/v1/all_cres` already returns.

### Pagination is opt-in — and why

Four things read this endpoint as *the complete list of roots*:

- `cre_main.download_cre_from_upstream` — the CLI import iterates `data["data"]` from opencre.org
- `DataProvider.rebuildDataTree` — the Explorer builds its whole tree from it
- the Browse page (`browseRootCres.tsx`)
- the `list_root_cres` MCP tool

If the endpoint paged by default, an upstream import would silently stop at the first 20 roots. So without `page`/`per_page` the response is byte-for-byte what it was; with either present it pages like `all_cres`. The spec description says so, and the handler comment names the consumers so the next person knows why.

```
GET /rest/v1/root_cres                      -> {"data": [...all roots...]}          (unchanged)
GET /rest/v1/root_cres?per_page=10&page=2   -> {"data": [...10...], "page": 2, "total_pages": 3}
GET /rest/v1/root_cres?per_page=1000        -> capped to 100
```

The root-CRE query is factored into `_root_cres_query()` and shared by `get_root_cres` and the new `get_root_cres_with_pagination`, so both return the same set in the same order.

### The other half of the issue is already done

The `all_cres` `per_page` cap the issue also asks for landed in 694f61e5 (`per_page = min(per_page, MAX_ITEMS_PER_PAGE)`, with `test_all_cres_caps_per_page` and the integration test). I did not touch it; this PR covers the `root_cres` half, which was still open.

### Tests

Written first, run before the implementation, and they failed on exactly the missing pieces (`'Node_collection' object has no attribute 'get_root_cres_with_pagination'`; `10 != 25` for `per_page=10`).

- `test_find_root_cres_paginates_only_when_asked` — no params: original shape, paginated method never called
- `test_find_root_cres_caps_per_page` — `per_page=1000` reaches the DB as 100
- `test_find_root_cres_pagination_integration` — real DB, 25 roots plus one contained child: pages 1 and 3 have 10 and 5, three pages cover all 25 with no repeats, the child never appears, `page` alone uses `ITEMS_PER_PAGE`, a page past the end is 404 rather than everything
- `test_get_root_cres_with_pagination` — collection level: pages equal slices of `get_root_cres()`, a page past the end is empty

`docs/api/openapi.yaml` is regenerated with `scripts/generate_openapi.py`, not hand-edited, and the guardrail passes. Because MCP input schemas derive from the spec, `list_root_cres` gains `page` and `per_page` with no catalog change (`format` still omitted).

### Checks

| | |
|---|---|
| `unittest discover` | 1013 tests; the only failures are the 8 `auth_routes_test` cases, identical on `main` with the same environment (they need `NO_LOGIN` unset, and pass then) |
| `make openapi-guardrail` | documented views / freshness / validity / route coverage all OK |
| `black` | clean |
| `mypy` (Makefile flags) on the four changed production files | 0 errors before and after |


kunalKumar-13:
Re-checked this against `main` after the recent `db.py` work (#1100, #1101, #1102) — it is still current, not another #850.

The reason it survived is that it was already written against the batch helpers rather than around them: `get_root_cres` calls `self._hydrate_cres_batch(...)` here, and `get_root_cres_with_pagination` does the same, so #1100's `_hydrate_nodes_batch` / `_hydrate_cres_batch` rework did not cut across it. All three anchors (`get_root_cres`, `_hydrate_cres_batch`, `all_cres_with_pagination`) are still on `main` at the shapes this builds on, and GitHub reports the PR `MERGEABLE`.

The shape, briefly, since it has been a couple of weeks:

- `_root_cres_query()` holds the query; `get_root_cres()` keeps its exact current behaviour on top of it.
- `.order_by(CRE.external_id, CRE.id)` — without a total order the database may return rows in any order, so a CRE could appear on two pages while another is never returned at all.
- `get_root_cres_with_pagination()` mirrors `all_cres_with_pagination` deliberately, including `error_out=False`, so a page past the end is empty rather than a 500.
- Unpaginated `/rest/v1/root_cres` is unchanged; pagination only applies when a `page` parameter is sent.

Happy to rebase if you'd rather take it after anything else in flight.

PR #1060, opened 2026-09-08
fix(api): validate and cap REST pagination parameters

kunalKumar-13:
Pagination on the documented REST API is unvalidated and, on the node endpoints, uncapped. Independent of #1059, which does the `root_cres` half of #847 — no overlapping lines.

### 1. A client typo is a 500, live right now

```
$ curl -o /dev/null -w '%{http_code}\n' 'https://opencre.org/rest/v1/all_cres?page=abc'
500
$ curl -o /dev/null -w '%{http_code}\n' 'https://opencre.org/rest/v1/standard/ASVS?page=abc'
500
```

`int(request.args.get("page"))` raises `ValueError`, nothing handles it, and Flask turns it into a server error. Now a 400, the way `text_search` reports a bad `text` parameter after #867. A value of zero or less still falls back to the default, exactly as before.

This affects **seven** documented endpoints, because `find_node_by_name` serves six of them:

```
/rest/v1/all_cres
/rest/v1/standard/{name}
/rest/v1/{ntype}/{name}
/rest/v1/{ntype}/{name}/section/{section}
/rest/v1/{ntype}/{name}/section/{section}/subsection/{subsection}
/rest/v1/{ntype}/{name}/sectionid/{sectionID}
/rest/v1/{ntype}/{name}/sectionid/{sectionID}/subsection/{subsection}
```

### 2. The node endpoints have no upper bound at all

`items_per_page` was read from the query string and handed to `paginate()` with nothing in between — no `MAX_ITEMS_PER_PAGE`, no clamp. With 250 nodes in the database:

| request | standards returned |
|---|---|
| `?items_per_page=999999` | **250** — the whole table |
| `?items_per_page=1000` | **250** |
| `?items_per_page=10` | 10 |

That is the unbounded response #847 describes, on the node endpoints rather than the CRE ones. Capped at `MAX_ITEMS_PER_PAGE` now, the same as `all_cres`.

### 3. `all_cres` pages are not stable

`all_cres_with_pagination` paginated `session.query(CRE)` with no `ORDER BY`. SQL does not define row order without one, so page boundaries can move between requests: a CRE can come back on two pages while another is never returned. Ordering is now `external_id` — the order the catalogue reads in — with the primary key breaking ties so the order is total.

**Honest caveat:** this one does not reproduce on SQLite, which hands back a small table in insertion order. That test is a regression guard, not a reproduction. It matters on Postgres, which is what production runs.

### Tests

Written first and run before each fix, and they failed on the real errors — `ValueError: invalid literal for int() with base 10: 'abc'` for the parsing, `AssertionError: 100 != 150` for the missing cap.

- `test_all_cres_rejects_malformed_pagination`
- `test_all_cres_pages_deterministically`
- `test_find_node_by_name_rejects_malformed_pagination` — covers `/standard/{name}`, `/section/`, `/sectionid/`
- `test_find_node_by_name_caps_items_per_page` — 150 nodes, `?items_per_page=999999`, asserts the cap and that a request under it is still honoured

### Checks

| | |
|---|---|
| `unittest discover` | **1013 tests, OK** |
| `make openapi-guardrail` | all four checks OK (behaviour only, no spec change) |
| `black` | clean |
| `mypy` with the Makefile flags on both changed files | 0 errors, same as `main` |


kunalKumar-13:
Still applies — I re-checked the current `main` rather than assuming.

The unguarded casts are all three still there:

```
web_main.py:258   if request.args.get("page") is not None and int(request.args.get("page")) > 0:
web_main.py:1501  if request.args.get("page") is not None and int(request.args.get("page")) > 0:
web_main.py:1508      per_page = int(request.args.get("per_page"))
```

`int()` raises `ValueError` there with nothing to catch it, so `?page=abc` — a client typo — comes back as a 500 rather than a 400.

The cap is the part worth a second look. `MAX_ITEMS_PER_PAGE = 100` already exists at line 57 and `all_cres` already applies it at line 1509, but `find_node_by_name` does not: `items_per_page` is read straight from the query string and handed to the query. So one request can ask for the whole table from that endpoint today, while the neighbouring endpoint is capped. This PR just makes the two consistent.

Behaviour that deliberately does not change: a `page` or `per_page` of zero or less still falls back to the default rather than erroring, which is what the current code does.

Rebased or split if either would help.

PR #1086, opened 2026-09-08
docs(oie-apis): fix relative link to Pane of Glass RFC

ajmani-x:
### What

`docs/owasp-graph/apis/README.md` links the Pane of Glass RFC as `../designs/owasp-pane-of-glass.md`. From `docs/owasp-graph/apis/` thatn resolves to `docs/owasp-graph/designs/owasp-pane-of-glass.md`, which does
not exist, so the link 404s.

The file is at `docs/designs/owasp-pane-of-glass.md` — one directory higher. The sibling `docs/owasp-graph/README.md` already links it correctly with `../designs/...` because it sits one level up.

### Fix

Add the missing `../` so the link points at `../../designs/owasp-pane-of-glass.md`.

### Validation

Checked locally that the corrected path resolves to the existing file and that no other doc references the broken path.


ajmani-x:
Hi, just checking in on this whenever you get a chance. Happy to make any changes if needed. Thanks!

PR #1092, opened 2026-09-12
fix(oscal): use the correct sectionID in list export

Devansh-567:
### PR fix

While looking through the OSCAL export code, I noticed that list_to_oscal() was using doc.section when creating the sectionID property.

This meant that the exported OSCAL data could contain the section text as both section and sectionID, while the actual sectionID value was never exported.

The issue is a small field mix-up here:
name="sectionID",
value="".join(doc.section.splitlines()).strip(),

This should use doc.sectionID instead.

### What changed

* Updated list_to_oscal() to use doc.sectionID for the sectionID OSCAL property.
* Fixed test_list_to_oscal so it actually tests list_to_oscal() nstead of document_to_oscal().
* Made the test data use different values for section and sectionID, so this kind of mix-up will be caught in the future.

### Testing

I verified that the new test fails with the old implementation and passes with the fix.

The full test suite passes with:
* 998 passed, 6 skipped, 0 failures


Devansh-567:
Hi @northdpole, are there any active reviewers?

PR #1095, opened 2026-09-12
fix(db): sanitize export filenames for Windows

### LINKED ISSUE #1094: db.export() filenames keep ':' and other reserved characters - export fails on Windows
While looking at db.export() I noticed it names every file after Document.id. For standards and tools that id is just the fields joined with colons (cre_defs.py:455), and the cleanup only strips slashes, spaces and quotes. Windows doesn't allow ':' in filenames, so export() throws OSError: [Errno 22] and writes nothing.

To reproduce, run test_export on Windows. It dies on the first write with a name like `Unlinked:Unlinked:4.5.6:https:--example.com.yaml`.

The fix is to sanitize the id before using it as a filename, covering the characters Windows rejects (<>:"/\|?*) and trailing dots/spaces. The `elif hasattr(doc, "sectionID")` branch next to it can be dropped too, every document has an id and it had the same problem.

Vishakha7-Kumari:
Fixes #1094.

While looking at the export code, I noticed it names each file after Document.id. For standards and tools that id is just the fields joined with colons, and the cleanup only strips slashes, spaces and quotes.

issue is here:

title = (
    doc.id.replace("/", "-")
    .replace(" ", "_")
    .replace('"', "")
    .replace("'", "")
    + ".yaml"
)

Windows doesn't allow ':' in filenames, so export() throws OSError: [Errno 22] and writes nothing.

I added safe_filename() to strip the characters Windows rejects and routed export() through it. I also dropped the leftover elif branch for sectionID, every document has an id and that branch had the same problem.

Testing
test_export fails on main and passes now. I added test_export_filenames_are_safe for a standard with colons in its sectionID.

The full suite runs 1010 tests. The 6 failures are Windows-only and fail on main too, nothing to do with this change.


northdpole (approved):
Line-by-line:

`safe_filename` — correct Win32 reserved set (`<>:"/\|?*` + controls). Space→`_` and trailing `.` / space strip match NTFS rules. `re` is already imported in `db.py`. Quotes now become `-` instead of being deleted; that is still a legal filename.

`export()` — using `doc.id` only is right. Every `CRE`/`Node`/`Standard`/`Tool` has `id`; the old `sectionID` branch was dead. YAML identity is still `todict()` inside the file, not the path. Nothing in CLI `--export` (CSV) or spreadsheet export reads these names.

`continue` after `logger.fatal` — safer than writing a broken title. Empty/`None` id is theoretical for this dump.

Test — colon-heavy Standard id is the actual Windows failure mode. The loop does not assert a specific file exists; acceptable for this dump helper.

LGTM. Rebase-merging.

PR #1101, opened 2026-09-13
Add standalone JSON logging (cre_logging)

northdpole:
## Summary
- Add top-level `cre_logging` so every module can `from cre_logging import get_logger` **without importing Flask/`application.__init__`**.
- Always emit one JSON object per stderr line, including `module`, `file`, `method`, and `level`.
- Replace per-module `logging.basicConfig()` / unused `pprint`, and send relevant `print` progress/errors through the logger. CLI machine output (`print(summary.to_json())`) stays on stdout.

Supersedes #856 (optional JSON + `application.utils.logging_config`, which still sits under the heavy application package).

## Test plan
- [x] `python -m unittest application.tests.cre_logging_test` (includes subprocess check that `import cre_logging` does not load `application` or `flask`)
- [x] mapping fixture / AISVS / cheatsheets parser tests
- [ ] CI `make lint` / `make mypy` / `make test`

Made with [Cursor](https://cursor.com)


northdpole:
Merging with admin bypass: Test, Lint, Frontend Lint, Librarian, Analyze (python), and both e2e jobs are green. CodeQL failed on pre-existing `web_main.py` alerts (path expression, clear-text secret, URL redirect, exception leak) — GitHub notes these can surface on large diffs even when not introduced by the PR. Logging change does not add those sinks.

